PR #1, opened 2016-08-24
make the root scope responsible for periodic reporting. make better m…

jra3:
This adds special logic to NewScope that creates a registry of scopes/subscopes and attaches a timer loop that will dump the stats from the scopes to the StatsReporter.

I decided to update the example to illustrate how this api is to be used, and to detach it from the cactus library (real implementations for statsd should live in some other project)


scope.go
```diff
@@ -23,12 +23,14 @@ package tally
 import (
 	"fmt"
 	"sync"
+	"time"
 
 	"github.com/facebookgo/clock"
 )
 
 type scope interface {
```
martin-mao: I don't see where this interface is used?

scope.go
```diff
@@ -103,6 +129,21 @@ func (s *standardScope) Report(r StatsReporter) {
 	// we do nothing for timers here because timers report directly to ths StatsReporter without buffering
 }
 
+// reportLoop is used by the root scope for periodic reporting
+func (s *standardScope) reportLoop(interval time.Duration) {
+	ticker := time.NewTicker(interval)
```
martin-mao: Might be a good idea to close a Scope so we don't leak go-routines once we don't need the scope any more (ie in testing etc)

scope.go
jra3: "scope" is included in the "Scope" interface. It contains the non-exported methods

example/main.go
```diff
@@ -0,0 +1,74 @@
...
+	"fmt"
+	"time"
+
+	"github.com/uber-go/tally"
+)
+
+type printStatsReporter struct{}
+
+// NewCactusStatsReporter returns a new StatsReporter
+func newPrintStatsReporter() tally.StatsReporter {
```
robskillington: Perhaps delete the comment or rename to reflect the name of the function "newPrintStatsReporter"

scope.go
jra3: Yeah, you have a point

robskillington: Agreed, a scope.Close() method would be ideal.

scope.go
```diff
@@ -103,6 +129,21 @@ func (s *standardScope) Report(r StatsReporter) {
...
+	for {
+		select {
+		case <-ticker.C:
+			s.registry.sm.Lock()
```
robskillington: Do you need a write lock here?  Can you get away with just a RLock() and RUnlock()?

scope.go
```diff
@@ -87,7 +84,36 @@ type standardScope struct {
...
+		tags = make(map[string]string)
+	}
+
+	scope := &standardScope{
+		prefix:   prefix,
+		tags:     tags,
+		reporter: reporter,
+
+		registry: &scopeRegistry{
+			subscopes: make([]Scope, 0, 5),
```
robskillington: Five is kind of a strange number, perhaps a bigger number like 4096 or just don't allocate anything upfront at all for the default allocator?

scope.go
jra3: So we are either writing with registry.add, or we are reading from reportLoop. reportLoop is only called from one goroutine, so there's no point in using an RWMutex.

scope.go
jra3: 5 is the best number. I've got all the best numbers

scope.go
```diff
@@ -21,12 +21,18 @@
...
+	"errors"
 	"fmt"
 	"sync"
+	"time"
 
 	"github.com/facebookgo/clock"
 )
 
+var (
+	errCloneNonRoot = errors.New("Called Close on non-root Scope")
```
robskillington: We could create a new interface called RootScope which inherits Scope to avoid this perhaps?

jra3: 👍

robskillington:
LGTM

martin-mao:
LGTM

PR #2, opened 2016-08-26
Add Flush() method to the StatsReporter interface

jra3:
It is useful to the implementer to get notified of when the Scope just finished a bunch of work. They might choose to do nothing... but It is probably the case that you will want to flush a batch of data in a single network operation after Report is done, and it's way easier to add a Report method than to reimplement Scope.

@prashantv

### DIFF
--- a/stats.go
+++ b/stats.go
@@ -31,6 +31,9 @@ type StatsReporter interface {
 	ReportCounter(name string, tags map[string]string, value int64)
 	ReportGauge(name string, tags map[string]string, value int64)
 	ReportTimer(name string, tags map[string]string, interval time.Duration)
+
+	// Flush is expected to be called by the Scope when it finishes a round or reporting
+	Flush()
 }
 
 type reportableMetric interface {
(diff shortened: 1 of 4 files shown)


stats.go
```diff
@@ -115,8 +118,13 @@ func (t *timer) Record(interval time.Duration) {
...
-func (r nullStatsReporter) ReportCounter(name string, tags map[string]string, value int64)          {}
-func (r nullStatsReporter) ReportGauge(name string, tags map[string]string, value int64)            {}
-func (r nullStatsReporter) ReportTimer(name string, tags map[string]string, interval time.Duration) {}
+func (r nullStatsReporter) ReportCounter(name string, tags map[string]string, value int64) {
+}
+func (r nullStatsReporter) ReportGauge(name string, tags map[string]string, value int64) {
+}
+func (r nullStatsReporter) ReportTimer(name string, tags map[string]string, interval time.Duration) {
+}
+func (r nullStatsReporter) Flush() {
```
prashantv: i kind of like the old formatting better where it just had the single line + aligned `{}`

prashantv:
lgtm

PR #5, opened 2016-08-31
Eliminate heap-allocation of Begin() timer function

jra3:
This cuts BenchmarkTimerInterval from ~86ns to ~41ns

Seems reasonable to do while we have zero adoption.

### DIFF
--- a/stats.go
+++ b/stats.go
@@ -54,10 +54,21 @@ type Gauge interface {
 	Update(int64)
 }
 
+type stopwatch struct {
+	startTime time.Time
+}
+
 // Timer is the interface for logging statsd-timer-type metrics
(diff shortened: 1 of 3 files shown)


stats.go
```diff
@@ -54,10 +54,21 @@ type Gauge interface {
 	Update(int64)
 }
 
+type stopwatch struct {
```
robskillington: Think this might have to be exported unfortunately, or else callers will not have access to the methods that return this type I believe.

robskillington: Also this could just be:
```go
type Stopwatch time.Time
```

robskillington:
LGTM

PR #7, opened 2016-10-19
Jra3/statsdreporter

jra3:
This is meant to serve as the most barebones implementation of a reporter than just wraps the existing statsd client implementation from github.com/cactus/go-statsd-client/statsd

The caller just needs to create a Statter pointing to their correct server:port and wrap it up.

### DIFF
--- a/statsd/reporter.go
+++ b/statsd/reporter.go
@@ -1,0 +1,63 @@
+package statsd
+
+import (
+	"sync"
+	"time"
+
+	"github.com/cactus/go-statsd-client/statsd"
+	"github.com/uber-go/tally"
+)
+
+type cactusStatsReporter struct {
+	statter  statsd.Statter
+	sm       sync.Mutex
+	scopes   []tally.Scope
+	interval time.Duration
+	quit     chan struct{}
+}
+
+// NewStatsdReporter wraps a statsd.Statter for use with tally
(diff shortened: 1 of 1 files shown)


glibsm:
fyi @sectioneight 
we'll probably want to include the result of this pr by the default for the oss scaffolds

statsd/reporter.go
```diff
@@ -0,0 +1,63 @@
...
+			for _, scope := range r.scopes {
+				scope.Report(r)
+			}
+		case <-r.quit:
+			return
+		}
+	}
+}
+
+func (r *cactusStatsReporter) registerScope(s tally.Scope) {
```
glibsm: what's this function for?

statsd/reporter.go
```diff
@@ -0,0 +1,63 @@
...
+	r.sm.Lock()
+	r.scopes = append(r.scopes, s)
+	r.sm.Unlock()
+}
+
+func (r *cactusStatsReporter) ReportCounter(name string, tags map[string]string, value int64) {
+	r.statter.Inc(name, value, 1.0)
```
glibsm: sample rate should probably be a parameter on NewStatsReporter

statsd/reporter.go
jra3: Like it says on the tin. A single reporter can have multiple scopes. E.g. creating a SubScope. This method is required to update the Reporter's internal bookkeeping.

PR #11, opened 2016-11-01
Allow access to the underlying reporter

glibsm:
Something that's required by uberfx for backwards compatibility support with old metrics system


robskillington:
Perhaps add a test that what it returns is the reporter it's instantiated with?

PR #12, opened 2016-11-02
Add access to tags and prefix on the scope

glibsm:
Similar to #11: allows to see which values were used during creation without having to remember them via other means

### DIFF
--- a/scope.go
+++ b/scope.go
@@ -57,6 +57,12 @@ type Scope interface {
 
 	// Reporter returns the underlying stats reporter which was used to initiate the Scope
 	Reporter() StatsReporter
+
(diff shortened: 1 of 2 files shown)


scope.go
```diff
@@ -273,6 +279,14 @@ func (s *standardScope) Reporter() StatsReporter {
 	return s.reporter
 }
 
+func (s *standardScope) Tags() map[string]string {
+	return s.tags
```
robskillington: Perhaps we should provide a copy of the tags map here so users of this library can't mangle the tags?  As it stands they could get the map then mutate it here which could cause all sorts of bad race conditions, etc.

PR #13, opened 2016-11-09
Adjust root types for a little more flexibility when implementing Scope and other types

robskillington:
cc @martin-mao @yurishkuro @xichen2020

### DIFF
--- a/scope.go
+++ b/scope.go
@@ -57,13 +57,11 @@ type scope struct {
 
 	cm sync.RWMutex
 	gm sync.RWMutex
-	hm sync.RWMutex
 	tm sync.RWMutex
 
-	counters   map[string]*counter
-	gauges     map[string]*gauge
-	histograms map[string]*histogram
-	timers     map[string]*timer
+	counters map[string]*counter
+	gauges   map[string]*gauge
+	timers   map[string]*timer
 }
 
 // NewRootScope creates a new Scope around a given stats reporter with the
@@ -106,10 +104,9 @@ func newRootScope(
 		registry: &scopeRegistry{},
 		quit:     make(chan struct{}),
 
-		counters:   make(map[string]*counter),
-		gauges:     make(map[string]*gauge),
-		histograms: make(map[string]*histogram),
-		timers:     make(map[string]*timer),
+		counters: make(map[string]*counter),
+		gauges:   make(map[string]*gauge),
+		timers:   make(map[string]*timer),
 	}
 
 	s.registry.add(s)
@@ -185,22 +182,6 @@ func (s *scope) Gauge(name string) Gauge {
 			s.gauges[name] = val
 		}
 		s.gm.Unlock()
-	}
-	return val
-}
-
-func (s *scope) Histogram(name string) Histogram {
-	s.hm.RLock()
-	val, ok := s.histograms[name]
-	s.hm.RUnlock()
-	if !ok {
-		s.hm.Lock()
-		val, ok = s.histograms[name]
-		if !ok {
-			val = newHistogram(s.fullyQualifiedName(name), s.tags, s.reporter)
-			s.histograms[name] = val
-		}
-		s.hm.Unlock()
 	}
 	return val
 }
@@ -291,16 +272,6 @@ func (s *scope) Snapshot() Snapshot {
 			}
 		}
 		ss.gm.RUnlock()
-		ss.hm.RLock()
-		for key, h := range ss.histograms {
-			name := ss.fullyQualifiedName(key)
-			snap.histograms[name] = &histogramSnapshot{
-				name:   name,
-				tags:   tags,
-				values: h.snapshot(),
-			}
-		}
-		ss.hm.RUnlock()
 		ss.tm.RLock()
 		for key, t := range ss.timers {
 			name := ss.fullyQualifiedName(key)
@@ -353,9 +324,6 @@ type Snapshot interface {
 	// Gauges returns a snapshot of gauge last values since last report execution
 	Gauges() map[string]GaugeSnapshot
 
-	// Histograms returns a snapshot of histogram values since last report execution
-	Histograms() map[string]HistogramSnapshot
-
 	// Timers returns a snapshot of timer values since last report execution
 	Timers() map[string]TimerSnapshot
 }
@@ -382,18 +350,6 @@ type GaugeSnapshot interface {
 
 	// Value returns the value
 	Value() float64
-}
-
-// HistogramSnapshot is a snapshot of a histogram
-type HistogramSnapshot interface {
-	// Name returns the name
-	Name() string
-
-	// Tags returns the tags
-	Tags() map[string]string
-
-	// Values returns the values
-	Values() []float64
 }
 
 // TimerSnapshot is a snapshot of a timer
@@ -425,18 +381,16 @@ func mergeRightTags(tagsLeft, tagsRight map[string]string) map[string]string {
 }
 
 type snapshot struct {
-	counters   map[string]CounterSnapshot
-	gauges     map[string]GaugeSnapshot
-	histograms map[string]HistogramSnapshot
-	timers     map[string]TimerSnapshot
+	counters map[string]CounterSnapshot
+	gauges   map[string]GaugeSnapshot
+	timers   map[string]TimerSnapshot
 }
 
 func newSnapshot() *snapshot {
 	return &snapshot{
-		counters:   make(map[string]CounterSnapshot),
-		gauges:     make(map[string]GaugeSnapshot),
-		histograms: make(map[string]HistogramSnapshot),
-		timers:     make(map[string]TimerSnapshot),
+		counters: make(map[string]CounterSnapshot),
+		gauges:   make(map[string]GaugeSnapshot),
+		timers:   make(map[string]TimerSnapshot),
 	}
 }
 
@@ -446,10 +400,6 @@ func (s *snapshot) Counters() map[string]CounterSnapshot {
 
 func (s *snapshot) Gauges() map[string]GaugeSnapshot {
 	return s.gauges
-}
-
-func (s *snapshot) Histograms() map[string]HistogramSnapshot {
-	return s.histograms
 }
 
 func (s *snapshot) Timers() map[string]TimerSnapshot {
@@ -490,24 +440,6 @@ func (s *gaugeSnapshot) Tags() map[string]string {
 
 func (s *gaugeSnapshot) Value() float64 {
 	return s.value
-}
-
-type histogramSnapshot struct {
-	name   string
-	tags   map[string]string
-	values []float64
-}
-
-func (s *histogramSnapshot) Name() string {
-	return s.name
-}
-
-func (s *histogramSnapshot) Tags() map[string]string {
-	return s.tags
-}
-
-func (s *histogramSnapshot) Values() []float64 {
-	return s.values
 }
 
 type timerSnapshot struct {
--- a/scope.go
+++ b/scope.go
@@ -72,29 +72,7 @@ func NewRootScope(
 	reporter StatsReporter,
 	interval time.Duration,
 ) (Scope, io.Closer) {
-	if tags == nil {
-		tags = make(map[string]string)
-	}
-
-	s := &scope{
-		prefix:   prefix,
-		tags:     tags,
-		reporter: reporter,
-
-		registry: &scopeRegistry{},
-		quit:     make(chan struct{}),
-
-		counters: make(map[string]*counter),
-		gauges:   make(map[string]*gauge),
-		timers:   make(map[string]*timer),
-	}
-
-	s.registry.add(s)
-
-	if interval > 0 {
-		go s.reportLoop(interval)
-	}
-
+	s := newRootScope(prefix, tags, reporter, interval)
 	return s, s
 }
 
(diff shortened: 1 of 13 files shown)


types.go
```diff
@@ -0,0 +1,160 @@
...
+
+	// SubScope returns a new child scope with the given name
+	SubScope(name string) Scope
+
+	// Capabilities returns a description of metrics reporting capabilities
+	Capabilities() Capabilities
+
+	// Snapshot returns a copy of all values since the last report execution,
+	// this is an expensive operation and should only be use for testing purposes
+	Snapshot() Snapshot
```
yurishkuro: this doesn't seem necessary, it's a feature of specific implementation like a "local backend", rather than the interface. Imaging a naive statsd implementation of Scope that emits each metric to UDP immediately, it wouldn't even be able to provide the snapshot functionality.

types.go
```diff
@@ -0,0 +1,160 @@
...
+}
+
+// RootScope is a scope that manages itself and other Scopes
+type RootScope interface {
```
yurishkuro: I don't think this needs to be a public interface. Imaging another naive implementation of Scope that simply logs each metric call - it has no need for "close". We've had a similar proposal in OpenTracing and decided against having Close method in the interface, because in practice you need to call Close() around the same place where you call New(), meaning you actually have access to the real implementation, not the interface.

In jaeger tracer we had the tracer struct itself private, so we could not return it from the New() method, but we handled it by returning `(opentracing.Tracer, io.Closer)`, so that in the main function people do

```
func main() {
    tracer, closer := jaeger.New()
    defer closer.Close()
   . . .
}
```

types.go
```diff
@@ -0,0 +1,160 @@
...
+// back to the timer's stop method at the end of the interval
+type StopwatchStart time.Time
+
+// Timer is the interface for logging statsd-timer-type metrics
+type Timer interface {
+	// Record Record a specific duration directly
+	Record(time.Duration)
+
+	// Start gives you back a specific point in time to report via Stop()
+	Start() StopwatchStart
```
yurishkuro: nit: I always found it odd that this is a part of the interface. It means every implementation needs to re-do it.  Why not just have a static method doing the same?
```
timer := scope.Timer(...)
...

func x() {
   sw := tally.Stopwatch(timer)
   defer sw.Stop()
   ...
}
```

types.go
```diff
@@ -0,0 +1,160 @@
...
+	// Update sets the gauges absolute value
+	Update(value int64)
+}
+
+// StopwatchStart is returned by a timer's start method, and should be passed
+// back to the timer's stop method at the end of the interval
+type StopwatchStart time.Time
+
+// Timer is the interface for logging statsd-timer-type metrics
+type Timer interface {
```
yurishkuro: Not sure how realistic it is to change this, but really the correct term for this metric is Histogram. When I joined Uber I think I asked Martin 2-3 times to explain the difference between timer and gauge - because there's no difference! The fact that timer also collects percentiles if a function of histogram, not inherently of the timer. As a counter example, I can imaging using a Gauge for the current length of in-memory queue. But what if I also want to get percentiles of that queue length? The only way to do it is with Timer, which doesn't make sense.

types.go
```diff
@@ -0,0 +1,160 @@
...
+
+	// Tags returns the tags
+	Tags() map[string]string
+
+	// Values returns the values
+	Values() []time.Duration
+}
+
+// StatsReporter is a backend for Scopes to report metrics to
+type StatsReporter interface {
```
yurishkuro: I suggest moving this to a different file, e.g. where it's actually used. The top level API only needs Scope and 3 metrics.

types.go
```diff
@@ -0,0 +1,160 @@
...
+	Gauge(name string) Gauge
+
+	// Timer returns the Timer object corresponding to the name
+	Timer(name string) Timer
+
+	// Tagged returns a new scope with the given tags
+	Tagged(tags map[string]string) Scope
+
+	// SubScope returns a new child scope with the given name
+	SubScope(name string) Scope
```
yurishkuro: these two methods, SubScope and Tagged, remind me of another difficulty I always had with go-common API w.r.t. proper separation of metric name and tags. Case in point - look at the [Metrics](https://github.com/uber/jaeger-client-go/blob/master/metrics.go#L31) struct in Jaeger client, especially the counters with tag `group` - I don't have a good feeling whether this should be a tag (dimension) or just a part of the metric name/namespace. It would be good to provide concrete examples when and what should be used in the comments. 

In particular, this is different from say go-kit where each metric type has a `WithTags()` method. That usage is more clear to me - you define a metric with a certain fixed name (and namespace), inject that metric into some service, and then the service has the option of adding dimensions.

This is a very subtle aspect of this API design, where a good design might clarify a lot of things, and a bad design makes the intended usage confusing.

types.go
robskillington: This is an issue however as you need to explicitly ask the Scope to report into the reporter.  The snapshot is simply to report what was buffered between reporting to the reporter from the scope.  It's purely a snapshot of what is buffered before reporting.  

In tests you need to ask the scope itself to give you a snapshot or have to wait until it's reported into a test reporter.  Waiting to be reported into a test reporter is a bad API.  It means you have to do some kind of wait/unsychronized routine.  Asking the Scope explicitly to report what data it has is a much better API.

The only way I can think of offering this without putting this on the scope is to cast up to a different type of scope _or_ creating a new type of constructor which takes no reporter but gives you the ability to take snapshots.  Perhaps I'll move to that model and add a different constructor that returns a `TestScope` with the `Snapshot() Snapshot` method.

types.go
robskillington: Sure, it's trivial to make this also return an `io.Closer`.  I still think it's a nicer API to get back a single thing rather than two references to the same thing, but if it simplifies the public API then I can get onboard with it.

types.go
robskillington: We do this to avoid allocations.  The method you propose introduces a required allocation as the stopwatch you return needs a reference to the timer.  

We can't really make it a struct either because then a struct becomes part of the public API.  I think this makes it restrictive to use and implement independently and we've seen import issues with this when one project imports the same dependency at different version (project x depends on v1 of dependency y, project x depends on dependency z which in turn depends on v2 of dependency y causing conflict struct definitions).

types.go
robskillington: I would prefer to keep all public APIs in single file for a project of this size.

types.go
robskillington: It seems poor to not allow users to prefix their metrics as they pass it down several levels deep in a project.  

Tagging should always be used for dimensions and subscope is for prefixing/namespacing your metrics.

Sharing one global namespace for all metric names definitely becomes messy.

types.go
yurishkuro: "you need to explicitly ask the Scope to report into the reporter" only in specific implementations.

types.go
yurishkuro: I disagree. The three distinct metric types are different by their intended use by the end user. Ultimately the only interaction the user has with them is giving them **a sequence of numbers**. The difference is what the user wants to do with the provided sequence of numbers by the end of the aggregation interval: to add them up and keep the sum (Counter), to keep the latest value (Gauge), or to build a distribution/percentiles (Histogram). None of it has anything to do with timers. Timer is just the interpretation of the values in some numeric sequences, a queue length could be another interpretation.

robskillington: That is the implementation of timers emitted at Uber however.  We can provide yet another type of metric that is intended to be evaluated as a histogram if you like.  

However users of a high level API should be able to also emit and collect raw timers if they should like.

types.go
yurishkuro:
```go
type Stopwatch struct {
    startTime time.Time
    timer tally.Timer
}

func GetSopwatch(timer tally.Timer) Stopwatch {
    return Stopwatch{time.Now(), timer}
}

func (sw Stopwatch) Stop() {
    sw.timer.Record(time.Since(sw.startTime).toMilliseconds())
}
```

no allocations. You can also do it with a private struct and expose Stopwatch interface, but I don't see much value in it given how simple the logic is.

robskillington: Once you return the stopwatch as an interface you cause the escape analysis of Go to allocate the stopwatch struct on the heap.  This causes allocation on the heap.

The only reason to avoid to returning just a plain exported struct is again the dependency management hell I described.

types.go
yurishkuro: I see timers as a consumer-side interpretation of a Histogram-style metric. The actual metrics backend binding to tally interface does not need to know that interpretation. To make things easier for Uber users we could have a static factory

```go
func NewTimer(scope Scope, name string, tags map[string]string) Timer {
    return &Timer{scope.Histogram(name, tags)} // could probably be an alias to Histogram
}
```

The Timer interface will expose the same notion of the timer as before, including  `Record(time.Duration)` which implies convention on the precision of the time values when they are translated into unit-less Historgam values

types.go
robskillington: I'll move it out to `reporter.go`

stats.go
```diff
@@ -21,71 +21,83 @@
...
-type StatsReporter interface {
-	ReportCounter(name string, tags map[string]string, value int64)
-	ReportGauge(name string, tags map[string]string, value int64)
-	ReportTimer(name string, tags map[string]string, interval time.Duration)
+var (
+	capabilitiesNone = &capabilities{
+		reporting: false,
+		tagging:   false,
+	}
+	capabilitiesNoTagging = &capabilities{
```
martin-mao: Nit: Maybe the naming here should indicate that it reports too?

stats.go
```diff
@@ -103,17 +115,25 @@ func (c *counter) report(name string, tags map[string]string, r StatsReporter) {
 	r.ReportCounter(name, tags, curr-prev)
```
martin-mao: Issue for line 110: why are we not loading c.prev atomically?

scope.go
```diff
@@ -162,118 +154,149 @@ func (s *standardScope) reportLoop(interval time.Duration) {
...
+			tags[k] = v
+		}
+
+		ss.cm.RLock()
+		for key, c := range ss.counters {
+			name := ss.fullyQualifiedName(key)
+			snap.counters[name] = &counterSnapshot{
+				name:  name,
+				tags:  tags,
+				value: c.snapshot(),
```
martin-mao: snapshot doesn't update the prev value, so for counters, snapshot is always going to be ever increasing. Is this the expected behavior or do we want snapshot to only have delta between snapshot calls?

types.go
```diff
@@ -0,0 +1,82 @@
...
+	Inc(delta int64)
+}
+
+// Gauge is the interface for logging statsd-gauge-type metrics
+type Gauge interface {
+	// Update sets the gauges absolute value
+	Update(value float64)
+}
+
+// Timer is the interface for logging statsd-timer-type metrics
```
yurishkuro: btw, just found a real example for the histogram argument in Jaeger code:

```go
	// to get accurate percentiles on the span size we emit it as a timer
	s.metrics.spanSizeBytes.Record(time.Duration(len(thriftSpan)) * time.Millisecond)
```

i.e. we have to fake a numeric value to be a timer

scope.go
robskillington: Snapshot is only the delta between reports, correct.

stats.go
robskillington: Sure thing, fixed.

stats.go
robskillington: Yeah, this looks bad.  Making a change to to load this atomically.

robskillington: Fixed.

robskillington:
Removing histogram until we have an implementation end to end that can support this type or at least another reporter that has support for this type.  The open source cactus statsd client has no support for histogram float value emission.

scope.go
```diff
@@ -283,4 +306,156 @@ func (s *standardScope) fullyQualifiedName(name string) string {
...
+	// Tags returns the tags
+	Tags() map[string]string
+
+	// Values returns the values
+	Values() []time.Duration
+}
+
+// mergeRightTags merges 2 sets of tags with the tags from tagsRight overriding values from tagsLeft
+func mergeRightTags(tagsLeft, tagsRight map[string]string) map[string]string {
+	if tagsLeft == nil && tagsRight == nil {
```
martin-mao: Nit: Any thoughts on checking left and right being nil to avoid creating new map unnecessarily in some cases?

stats.go
```diff
@@ -95,23 +71,67 @@ func (c *counter) Inc(v int64) {
...
 	}
 	atomic.StoreInt64(&c.prev, curr)
 	r.ReportCounter(name, tags, curr-prev)
 }
 
-func (g *gauge) Update(v int64) {
-	atomic.StoreInt64(&g.curr, v)
-	atomic.StoreInt64(&g.updated, 1)
+func (c *counter) snapshot() int64 {
+	return atomic.LoadInt64(&c.curr) - atomic.LoadInt64(&c.prev)
```
martin-mao: Don't we need to reset prev here much like report? Otherwise we'll snapshot this every second and the count will keep increasing forever, instead of giving you a delta between snapshots.

stats.go
```diff
@@ -126,12 +146,49 @@ func (t *timer) Record(interval time.Duration) {
...
 }
 
+func (t *timer) snapshot() []time.Duration {
+	t.unreported.RLock()
+	snap := make([]time.Duration, len(t.unreported.values))
+	for i := range t.unreported.values {
+		snap[i] = t.unreported.values[i]
+	}
+	t.unreported.RUnlock()
+	return snap
```
martin-mao: Again, do we just want to have more and more timers building up between snapshots?

robskillington: Yeah - this is only between reports and only accessible on a `TestScope`

martin-mao:
Ignore my comments, didn't realized snapshot was just for TestScope purposes. LGTM.

scope.go
robskillington: We should do this optimization probably, preferably not with this PR however.

PR #14, opened 2016-11-11
Stop pinning dependencies in glide.yaml

billf:
As part of glide's dependency resolution, consumers of tally will have to consider these specific versions. That's fine for semvers, but the specific revisions are too strict.


robskillington:
We actually want to be very specific which version we depend on to avoid running `glide update` in an application repository and suddenly changing actual semantics of underlying repositories by accident.

We are open to making these semvers, but definitely not master.

billf:
even for `testImports`?

robskillington:
There's probably less harm with `testImports`, having said that when you run `glide update` and things change other than what you're explicitly trying to update - now you have to investigate what changed and why.

The idea here is ensuring reproducible builds and _also_ reproducible tests.

PR #15, opened 2016-11-22
Implement Stopwatch as struct for nicer elapsed time measurement.

anakryiko:
This makes timer.Stop() API a bit nicer to use. Plus now you have ability to just pass Stopwatch to some other function without passing timer as well.


anakryiko:
In terms of performance both solutions are effectively equivalent:

BEFORE: 
BenchmarkTimerInterval-8        500000000               32.0 ns/op

AFTER:
BenchmarkTimerInterval-8        500000000               34.5 ns/op

anakryiko:
@robskillington

robskillington:
Thanks @anakryiko - I'm ok returning a struct, interface would've caused allocations, as a struct though it will live on the stack so I'm ok with this.

PR #22, opened 2017-01-20
Add CachedReporter interface that supports pre allocation.

Raynos:
This PR adds a new method `NewCachedRootScope` and a new
`CachedReporter` interface.

This `CachedReporter` interface allows for onetime allocation
of counters, gauges & timers. This allows for the CachedReporter
to cached the tags map and do pre-computation.

The actual runtime interface after all the scopes & metrics are
allocated at startup just involves `ReportCount(int64)` and friends.

This enables a reporter backend to implement a more involved BUT
more performant interface.

I've got a benchmark using an internal reporter where I was measuring the
overhead of adding 3 timers to a HTTP server.

 - master, 3 tally.Timers per request ~= 27% cpu overhead & 50 allocs per timer
 - this branch, 3 tally.Timers per request ~= 11% cpu overhead & 25 allocs per timer

This is a 2x improvement and allows us to have more timers in our
applications.

r: @robskillington @martin-mao
cc: @Matt-Esch @zhenghuiwang

### DIFF
--- a/reporter.go
+++ b/reporter.go
@@ -39,3 +39,37 @@ type StatsReporter interface {
 	// Flush is expected to be called by a Scope when it completes a round or reporting
 	Flush()
 }
+
+// CachedStatsReporter is a backend for Scopes that pre allocates all
+// counter, gauges & timers. This is harder to implement but more performant
+type CachedStatsReporter interface {
+	// AllocateCounter pre allocates a counter data structure with name & tags.
+	AllocateCounter(name string, tags map[string]string) CachedCount
+
+	// AllocateGauge pre allocates a gauge data structure with name & tags.
+	AllocateGauge(name string, tags map[string]string) CachedGauge
+
+	// AllocateTimer pre allocates a timer data structure with name & tags.
+	AllocateTimer(name string, tags map[string]string) CachedTimer
+
+	// CachedCapabilities returns a description of metrics reporting capabilities
+	CachedCapabilities() Capabilities
+
+	// CachedFlush is expected to be called by a Scope when it completes a round or reporting
+	CachedFlush()
+}
+
+// CachedCount interface for reporting an individual counter
+type CachedCount interface {
+	ReportCount(value int64)
+}
+
+// CachedGauge interface for reporting an individual gauge
+type CachedGauge interface {
+	ReportGauge(value float64)
+}
+
+// CachedTimer interface for reporting an individual timer
+type CachedTimer interface {
+	ReportTimer(interval time.Duration)
+}
--- a/reporter.go
+++ b/reporter.go
@@ -22,14 +22,15 @@ package tally
 
 import "time"
 
-type baseStatsReporter interface {
+// BaseStatsReporter implements the shared reporter methods
+type BaseStatsReporter interface {
 	Capabilities() Capabilities
 	Flush()
 }
 
 // StatsReporter is a backend for Scopes to report metrics to
 type StatsReporter interface {
-	baseStatsReporter
+	BaseStatsReporter
 
 	// ReportCounter reports a counter value
 	ReportCounter(name string, tags map[string]string, value int64)
@@ -44,7 +45,7 @@ type StatsReporter interface {
 // CachedStatsReporter is a backend for Scopes that pre allocates all
 // counter, gauges & timers. This is harder to implement but more performant
 type CachedStatsReporter interface {
-	baseStatsReporter
+	BaseStatsReporter
 
 	// AllocateCounter pre allocates a counter data structure with name & tags.
 	AllocateCounter(name string, tags map[string]string) CachedCount
--- a/reporter.go
+++ b/reporter.go
@@ -22,8 +22,15 @@ package tally
 
 import "time"
 
+type baseStatsReporter interface {
+	Capabilities() Capabilities
+	Flush()
+}
+
 // StatsReporter is a backend for Scopes to report metrics to
 type StatsReporter interface {
+	baseStatsReporter
+
 	// ReportCounter reports a counter value
 	ReportCounter(name string, tags map[string]string, value int64)
 
@@ -32,17 +39,13 @@ type StatsReporter interface {
 
 	// ReportTimer reports a timer value
 	ReportTimer(name string, tags map[string]string, interval time.Duration)
-
-	// Capabilities returns a description of metrics reporting capabilities
-	Capabilities() Capabilities
-
-	// Flush is expected to be called by a Scope when it completes a round or reporting
-	Flush()
 }
 
 // CachedStatsReporter is a backend for Scopes that pre allocates all
 // counter, gauges & timers. This is harder to implement but more performant
 type CachedStatsReporter interface {
+	baseStatsReporter
+
 	// AllocateCounter pre allocates a counter data structure with name & tags.
 	AllocateCounter(name string, tags map[string]string) CachedCount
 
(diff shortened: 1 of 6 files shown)


reporter.go
```diff
@@ -39,3 +39,37 @@ type StatsReporter interface {
...
+	// AllocateGauge pre allocates a gauge data structure with name & tags.
+	AllocateGauge(name string, tags map[string]string) CachedGauge
+
+	// AllocateTimer pre allocates a timer data structure with name & tags.
+	AllocateTimer(name string, tags map[string]string) CachedTimer
+
+	// CachedCapabilities returns a description of metrics reporting capabilities
+	CachedCapabilities() Capabilities
+
+	// CachedFlush is expected to be called by a Scope when it completes a round or reporting
```
robskillington: Might be worth fixing the doc string here, the "or" is meant to be "of".  Worth fixing on line 39 too I should think.

I think it might be worth renaming `CachedCapabilities` and `CachedFlush` to just `Capabilities` and `Flush` and then creating a new internal only interface:

```go
type baseStatsReporter interface {
  Capabilities() Capabilities()
  Flush()
}
```

That way you can add a field to the scope:
```go
  reporter StatsReporter   // existing
  cachedReporter CachedStatsReporter // existing
  baseReporter baseStatsReporter // new
```

And when calling capabilities or flush:

```go
// capabilities
s.baseReporter.Capabilities()
// flush
s.baseReporter.Flush()
```

Just to avoid the if/elses there.

scope.go
```diff
@@ -305,6 +376,9 @@ func (s *scope) Close() error {
 	if closer, ok := s.reporter.(io.Closer); ok {
 		return closer.Close()
 	}
+	if closer, ok := s.cachedReporter.(io.Closer); ok {
+		return closer.Close()
+	}
```
robskillington: If we were using the baseStatsReporter approach this could also just be:

```go
 	if closer, ok := s.baseReporter.(io.Closer); ok {
  		return closer.Close()		  		return closer.Close()
  	}
```

stats.go
```diff
@@ -79,17 +80,30 @@ func (c *counter) report(name string, tags map[string]string, r StatsReporter) {
...
 }
 
+func (c *counter) cachedReport() {
+	curr := atomic.LoadInt64(&c.curr)
+
+	prev := atomic.LoadInt64(&c.prev)
+	if prev == curr {
+		return
+	}
+	atomic.StoreInt64(&c.prev, curr)
```
robskillington: Might be best to avoid repeating this code by adding a new method that takes the value?

```go
func (c *counter) value() int64 {
	curr := atomic.LoadInt64(&c.curr)

	prev := atomic.LoadInt64(&c.prev)
	if prev == curr {
		return 0
	}
    return curr - prev
}
```

stats.go
```diff
@@ -103,6 +117,14 @@ func (g *gauge) report(name string, tags map[string]string, r StatsReporter) {
 	}
 }
 
+func (g *gauge) cachedReport() {
+	if atomic.SwapUint64(&g.updated, 0) == 1 {
+		g.cachedGauge.ReportGauge(
+			math.Float64frombits(atomic.LoadUint64(&g.curr)),
```
robskillington: Maybe same for the gauge, something that just performs the `math.Float64frombits(atomic.LoadUint64(&g.curr))`.  The compiler will likely inline it so will stay fast but is more concise to read.

stats.go
```diff
@@ -139,7 +168,11 @@ func (t *timer) Start() Stopwatch {
 }
 
 func (t *timer) Record(interval time.Duration) {
-	t.reporter.ReportTimer(t.name, t.tags, interval)
+	if t.cachedTimer != nil {
+		t.cachedTimer.ReportTimer(interval)
+	} else {
+		t.reporter.ReportTimer(t.name, t.tags, interval)
+	}
```
robskillington: Might be worth binding this method to avoid the if/else jump at `Record` call time:

```go

type timer struct {
  // ...
  reportFn func(interval time.Duration)
}

func newTimer(/*...*/) { 
  t := &timer{/*...*/}
  if cachedTimer != nil {
    t.reportFn = t.cacheTimer.ReportTimer
  } else {
    t.reportFn = t.report
  }
}

func (t *timer) Record(interval time.Duration) {
  t.reportFn(interval)
}

func (t *timer) report(interval time.Duration) {
  t.reporter.ReportTimer(t.name, t.tags, interval)
}
```

scope.go
```diff
@@ -247,10 +315,13 @@ func (s *scope) SubScope(prefix string) Scope {
...
-	if s.reporter == nil {
-		return capabilitiesNone
+	if s.reporter != nil {
+		return s.reporter.Capabilities()
 	}
-	return s.reporter.Capabilities()
+	if s.cachedReporter != nil {
+		return s.cachedReporter.CachedCapabilities()
+	}
+	return capabilitiesNone
```
robskillington: If we were using the baseStatsReporter approach this could also just be:
```go
if s.baseReporter != nil [
  return s.baseReporter.Capabilities()
}
return capabilitiesNone
```

stats.go
Raynos: I tried this.

```
raynos at raynos-ThinkPad-T440p  
~/gocode/src/github.com/uber-go/tally on cached-reporter*
$ go test -bench BenchmarkTimerReport -run _NONE_ -cpuprofile cpuprofile.out -benchmem -benchtime 7s -cpu 2 .
BenchmarkTimerReport-2   	2000000000	         6.02 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	github.com/uber-go/tally	12.633s
raynos at raynos-ThinkPad-T440p  
~/gocode/src/github.com/uber-go/tally on cached-reporter*
$ go test -bench BenchmarkTimerReport -run _NONE_ -cpuprofile cpuprofile.out -benchmem -benchtime 7s -cpu 2 .
BenchmarkTimerReport-2   	1000000000	         9.24 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	github.com/uber-go/tally	10.196s
```

It looks like the inline if statement is faster then the bound function call. Adding an extra function call takes 9/10ns instead of 5/6ns.

stats.go
Raynos: Fixed.

stats.go
Raynos: Fixed.

reporter.go
Raynos: Fixed. However this means that the `Flush()` method name classes and a single struct can no longer implement both StatsReporter and CachedStatsReporter, but that's probably a good thing.

scope.go
Raynos: Fixed.

reporter.go
```diff
@@ -22,8 +22,15 @@ package tally
 
 import "time"
 
+type baseStatsReporter interface {
```
robskillington: Might be worth making this a public interface on second thoughts since its strange to export a public interface that extends a privately defined interface.

stats.go
robskillington: Interesting, sounds good.  Thanks for the breakdown 👍

robskillington:
Other than final comment LGTM

reporter.go
Raynos: Fixed.

Raynos:
@robskillington 

After I land this we should cut a proper version and use glide versioning in tally.

Since I'm going to be using uber-go/tally in prod, I'm happy to cut 1.0.0 and lock down the interface ;)

Alternatively we can version it as 0.1.0 if there are known things that need to change.

Raynos:
Ok, will wait until monday for landing, to catch more reviewers.

Raynos:
I've merged this and then started using semantic versioning in this repository, starting at v1.0.0

PR #24, opened 2017-01-30
Add M3 reporter backend to tally

Raynos:
This adds the M3 reporter client to tally as a first
class reporter that can be used.

Whilst open sourcing two breaking changes were added:

 - commonTags[env] is required and not pre-populated.
 - commonTags[host] is required and not pre-populated.

The logic to compute host & env is not open-sourcable so
users of the m3backend will need to pass env and host in
themselves.

Other then that this a straight up port & namespace refactor

r: @robskillington @martin-mao
cc: @Matt-Esch @zhenghuiwang @stevededalus


Raynos:
Thanks @mention-bot  :(

But actually, @robskillington  & @martin-mao are reviewers.

.travis.yml
```diff
@@ -12,6 +12,7 @@ cache:
   directories:
     - vendor
 install:
+  - npm i uber-licence -g
```
robskillington: Would avoiding `-g` be best for travis and just reference by path directly `./node_modules/.bin/...` be more worthwhile?

Raynos: Works for me.

m3/m3backend.go
```diff
@@ -0,0 +1,940 @@
...
+func NewMultiM3BackendWithProtocol(
+	hostPorts []string,
+	service string,
+	commonTags map[string]string,
+	includeHost bool,
+	maxQueueSize int,
+	maxPacketSizeBytes int32,
+	interval time.Duration,
+	protocol Protocol,
+) (BufferedBackend, error) {
```
robskillington: Hey let's turn these three constructors into a single constructor with just an exported option struct to make adding fields not a breaking change. e.g.

```go
backend, err := NewM3Backend(BackendOptions{
  HostPorts: []string{"127.0.0.1:9052"},
  // ..
})
```

PR #25, opened 2017-02-05
Return existing scopes when using Tagged and SubScope

robskillington:
cc @martin-mao

### DIFF
--- a/key_gen.go
+++ b/key_gen.go
@@ -1,0 +1,124 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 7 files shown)


key_gen.go
```diff
@@ -0,0 +1,124 @@
...
+	writeKeyValuesSortedByKeys(buf, keys, stringMap)
+
+	key := buf.String()
+	keyGenPool.release(buf, keys)
+	return key
+}
+
+// KeyForPrefixedStringMap generates a unique key for a
+// a prefix and a map string set combination.
+func KeyForPrefixedStringMap(
```
martin-mao: Can KeyForStringMap just call this with empty prefix and we do a check in this func?

robskillington: Can do, hopefully not much overhead.

PR #26, opened 2017-02-05
Add multi reporters to support multiple backends for a single scope

robskillington:
cc @martin-mao @akshayjshah

### DIFF
--- a/multi/reporter.go
+++ b/multi/reporter.go
@@ -106,7 +106,7 @@ func (r *multiCached) AllocateCounter(
 	name string,
 	tags map[string]string,
 ) tally.CachedCount {
-	var metrics []tally.CachedCount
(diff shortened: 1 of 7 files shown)


multi/reporter.go
```diff
@@ -0,0 +1,198 @@
...
+		multiBaseReporters: baseReporters,
+		reporters:          r,
+	}
+}
+
+func (r *multiCached) AllocateCounter(
+	name string,
+	tags map[string]string,
+) tally.CachedCount {
+	var metrics []tally.CachedCount
```
martin-mao: Nit: Make with len of reporters and use i in following loop to assign.

robskillington: Sure thing.

PR #28, opened 2017-02-07
Add native histogram support

robskillington:
This change adds a native `Histogram` type to tally.  It has the following interface:
```go
// Histogram is the interface for emitting histogram metrics
type Histogram interface {
	// RecordValue records a specific value directly.
	// Will use the configured value buckets for the histogram.
	RecordValue(value float64)

	// RecordDuration records a specific duration directly.
	// Will use the configured duration buckets for the histogram.
	RecordDuration(value time.Duration)

	// Start gives you a specific point in time to then record a duration.
	// Will use the configured duration buckets for the histogram.
	Start() Stopwatch
}
```

You can create it from a scope using the signature:

```go
type Scope interface {
	// ...

	// Histogram returns the Histogram object corresponding to the name.
	// To use default value and duration buckets configured for the scope
	// simply pass tally.DefaultBuckets.
	// You can use tally.ValueBuckets(x, y, ...) for value buckets.
	// You can use tally.DurationBuckets(x, y, ...) for duration buckets.
	// You can use tally.LinearValueBuckets(start, width, count) for linear values.
	// You can use tally.LinearDurationBuckets(start, width, count) for linear durations.
	// You can use tally.ExponentialValueBuckets(start, factor, count) for exponential values.
	// You can use tally.ExponentialDurationBuckets(start, factor, count) for exponential durations.
	Histogram(name string, buckets Buckets) Histogram

	// ...
}
```

Using helpers or inbuilt bucket types you can create a histogram like so:

```go
// use default value and duration buckets
h := scope.Histogram("foo", tally.DefaultBuckets) // default buckets, can pass nil too
h.RecordValue(42.42) // record float value using default value buckets
h.RecordDuration(42*time.Millisecond) // record duration value using default duration buckets

// use custom duration buckets 
h := scope.Histogram("foo", DurationBuckets{100*time.Millisecond, 200*time.Millisecond}) 
h.RecordDuration(150*time.Millisecond)

// use custom value buckets with linear helper
h := scope.Histogram("foo", LinearValueBuckets(1.5, 1.5, 3)) 
h.RecordValue(3.3)

// use custom duration buckets with exponential helper
h := scope.Histogram("foo", ExponentialDurationBuckets(5*time.Millisecond, 1.5, 8)) 
h.RecordDuration(8*time.Millisecond)
```

You can specify default buckets on a `Scope` using `DefaultBuckets`:

```go
// ScopeOptions is a set of options to construct a scope.
type ScopeOptions struct {
	// ...
	DefaultBuckets    []Bucket
	// ...
}
```

cc @Raynos @akshayjshah @martin-mao @xichen2020


robskillington:
Note: this is a WIP, the top level `tally` package passes all tests and compiles, the reporters do not currently.

Feedback before I go ahead with the reporter implementations would be good.  There were quite a few variations I tried on this but this was the cleanest.

akshayjshah:
1. What's the expected use case of `DurationBucket` and `ValueBucket`? Can I mix them in a single histogram? Perhaps the API would be simpler with just `DurationBuckets` and `ValueBuckets` (and the various helpers).
2. On the M3 side, what does it mean to record both values and durations in a single histogram? Are durations implicitly treated as an integer number of nanoseconds? Might be simpler to have `Scope.DurationHistogram` and `Scope.ValueHistogram` be completely separate.
3. Because the main exported types in Tally are interfaces (scope, counter, timer, gauge), adding this feature is a breaking change. While we're bumping the version to 2.0, might as well clean up any other APIs that need it. In particular, perhaps *all* types should be cached (i.e., remove the non-caching variants).

robskillington:
Re: 1. I actually did add the `Start() Stopwatch` to the `Histogram` type here:
```go
type Histogram interface {
        // ...

	// Start gives you a specific point in time to then record a duration.
	// Will use the configured duration buckets for the histogram.
	Start() Stopwatch
}
```

Re: 2. & 3. I am ok making them two different histogram types, that's fine.  On the M3 side it will cause issues if both are emitted at the same time with different buckets as the buckets will overlap.  For prometheus it will be "ok" as whatever is first emitted, durations or buckets, will set the total buckets for the entire histogram and all values will just flow into that.  Still not great though and depending which code path you go down first determines which buckets are used.

Re: 4. Yeah this will require us to issue a 2.0 - which is fine as long as users are using semver.  We can remove the non-cached reporters but it does put a large onus on implementing your own reporter.  I guess if we feel ok with that, I'm ok to get rid of the non-caching variants.

akshayjshah:
Ah, you're right - missed the `Start` method.

Two types seems less error-prone to me - if you're okay with that, let's separate them.

If it's okay with you, dropping the uncached reports and reducing the surface area here seems like a win. I haven't looked at the interfaces in detail, but could we ship a

```
func NopCache(UncachedReporter) CachedReporter
```

if we ever want to support less-performant reporters?

histogram.go
```diff
@@ -0,0 +1,183 @@
...
+	for i, v := range values {
+		buckets[i] = DurationBucket(v)
+	}
+	return buckets
+}
+
+// LinearValueBuckets creates a set of linear value buckets.
+func LinearValueBuckets(start, width float64, count int) []Bucket {
+	if count <= 0 {
+		panic("count needs to be > 0")
```
xichen2020: Should we return an error instead of panicing?

histogram.go
```diff
@@ -0,0 +1,183 @@
...
+	for i := range buckets {
+		buckets[i] = ValueBucket(start + (float64(i) * width))
+	}
+	return buckets
+}
+
+// LinearDurationBuckets creates a set of linear duration buckets.
+func LinearDurationBuckets(start, width time.Duration, count int) []Bucket {
+	if count <= 0 {
+		panic("count needs to be > 0")
```
xichen2020: Same here

scope.go
```diff
@@ -269,25 +261,68 @@ func (s *scope) Timer(name string) Timer {
...
+	}
+
+	key := name +
+		Buckets(valueBuckets).String() +
+		Buckets(durationBuckets).String()
+
+	s.tm.RLock()
+	val, ok := s.histograms[key]
+	s.tm.RUnlock()
+	if !ok {
```
xichen2020: Could reduce nesting here by doing `if ok {...}` first

scope.go
```diff
@@ -269,25 +261,68 @@ func (s *scope) Timer(name string) Timer {
...
+		s.tm.Lock()
+		val, ok = s.histograms[key]
+		if !ok {
```
xichen2020: Could reduce nesting here

stats.go
```diff
@@ -193,6 +206,19 @@ func (t *timer) snapshot() []time.Duration {
...
 }
 
+type timerStopwatch struct {
+	start   time.Time
+	timer   *timer
+	stopped int32
+}
+
+func (s timerStopwatch) Stop() {
+	if atomic.AddInt32(&s.stopped, 1) == 1 {
```
xichen2020: `atomic.CompareAndSwapInt32(&s.stopped, 0, 1)` might be easier to understand

stats.go
```diff
@@ -207,13 +233,74 @@ func (r *timerNoReporterSink) ReportTimer(name string, tags map[string]string, i
...
+		tags:            tags,
+		reporter:        r,
+		buckets:         buckets,
+		cachedHistogram: cachedHistogram,
+	}
+}
+
+func (h *histogram) RecordValue(value float64) {
+	if h.cachedHistogram != nil {
+		h.cachedHistogram.ReportHistogramValue(value)
```
Raynos: Histograms can be aggregated. Tally aggregates counters & gauges inside tally itself instead of the reporter.

It might make sense to aggregate in tally and to report/flush in a tick loop instead of reporting every single histogram.

robskillington: Good call.

Yeah, what is tough is the Prometheus API doesn't have a way to report already aggregated values into the interface.

I can get around this however by doing a `for _, b := range buckets { for i := 0; i < samples; i++; { b.Observe(b.Value()) } }` however.  Will attempt this approach.

stats.go
robskillington: Sure thing.

robskillington:
@akshayjshah @Raynos @xichen2020 this PR is good to review.

m3/reporter_test.go
```diff
@@ -222,6 +222,88 @@ func TestReporterFinalFlush(t *testing.T) {
...
+		}[tag.GetTagName()], tag.GetTagValue())
+	}
+	require.True(t, counter.IsSetMetricValue())
+	val = counter.GetMetricValue()
+	require.True(t, val.IsSetCount())
+	require.False(t, val.IsSetGauge())
+	require.False(t, val.IsSetTimer())
+	count = val.GetCount()
+	require.True(t, count.IsSetI64Value())
+	require.EqualValues(t, int64(3), count.GetI64Value())
```
akshayjshah: It's pretty difficult to tell what's different between these two assertion blocks. Can we pull the common stuff out into a helper function?

m3/reporter_test.go
```diff
@@ -222,6 +222,88 @@ func TestReporterFinalFlush(t *testing.T) {
...
+		}[tag.GetTagName()], tag.GetTagValue())
+	}
+	require.True(t, counter.IsSetMetricValue())
+	val := counter.GetMetricValue()
+	require.True(t, val.IsSetCount())
+	require.False(t, val.IsSetGauge())
+	require.False(t, val.IsSetTimer())
+	count := val.GetCount()
+	require.True(t, count.IsSetI64Value())
+	require.EqualValues(t, int64(7), count.GetI64Value())
```
akshayjshah: Why use `EqualValues` instead of `Equal` here? You're already explicitly casting the constant to `int64`.

multi/reporter.go
```diff
@@ -167,13 +208,46 @@ func (m multiMetric) ReportTimer(interval time.Duration) {
...
+}
+
 type multiBaseReporters []tally.BaseStatsReporter
 
 func (r multiBaseReporters) Capabilities() tally.Capabilities {
 	c := &capabilities{reporting: true, tagging: true}
 	for _, r := range r {
 		c.reporting = c.reporting && r.Capabilities().Reporting()
 		c.tagging = c.tagging && r.Capabilities().Tagging()
+		c.histograms = c.tagging && r.Capabilities().Histograms()
```
akshayjshah: Should this be `c.histograms`, and should the seed value set `histograms` to true?

Does this have any test coverage?

reporter.go
```diff
@@ -22,39 +22,88 @@ package tally
...
+		samples int64,
+	)
+
+	// ReportHistogramDurationSamples reports histogram samples for a bucket
+	ReportHistogramDurationSamples(
+		name string,
+		tags map[string]string,
+		buckets Buckets,
+		bucketLowerBound,
+		bucketUpperBound time.Duration,
```
akshayjshah: Why do we have to pass the lower & upper bounds separately? Can't they be inferred from the buckets themselves? (Or be part of the buckets interface?)

scope.go
```diff
@@ -31,11 +31,29 @@ import (
...
+		100 * time.Millisecond,
+		200 * time.Millisecond,
+		300 * time.Millisecond,
+		400 * time.Millisecond,
+		500 * time.Millisecond,
+		600 * time.Millisecond,
+		800 * time.Millisecond,
+		1 * time.Second,
+		2 * time.Second,
+		5 * time.Second,
```
akshayjshah: :100: thanks for going through that long discussion about the defaults.

m3/reporter.go
```diff
@@ -243,6 +266,79 @@ func (r *reporter) AllocateTimer(
...
+	name string,
+	tags map[string]string,
+	buckets tally.Buckets,
+) tally.CachedHistogram {
+	var (
+		durations             = buckets.AsDurations()
+		lookupByValue         = buckets.AsValues()
+		lookupByDuration      = make([]int, buckets.Len())
+		cachedValueBuckets    []cachedMetric
+		cachedDurationBuckets []cachedMetric
```
robskillington: These are mutated only one time at histogram allocation time.

m3/reporter.go
```diff
@@ -243,6 +266,79 @@ func (r *reporter) AllocateTimer(
...
+
+		valueTags[idTagName] = fmt.Sprintf(bucketIDFmt, i)
+		valueTags[nameTagName] = fmt.Sprintf("%s-%s",
+			r.valueBucketString(pair.LowerBoundValue()),
+			r.valueBucketString(pair.UpperBoundValue()))
+
+		cachedValueBuckets = append(cachedValueBuckets,
+			r.allocateCounter(name, valueTags))
+
+		durationTags[idTagName] = fmt.Sprintf(bucketIDFmt, i)
```
robskillington: I can change this code to use the same bucket ID, they do have to use different names however.  Again this only happens at initial allocation time.

example/main.go
```diff
@@ -45,6 +45,30 @@ func (r *printStatsReporter) ReportTimer(name string, tags map[string]string, in
 	fmt.Printf("timer %s %s\n", name, interval.String())
 }
 
+func (r *printStatsReporter) ReportHistogramValueSamples(
+	name string,
+	tags map[string]string,
+	buckets tally.Buckets,
```
robskillington: Agreed, sure thing.

example/main.go
```diff
@@ -45,6 +45,30 @@ func (r *printStatsReporter) ReportTimer(name string, tags map[string]string, in
...
+	samples int64,
+) {
+	fmt.Printf("histogram %s bucket lower %f upper %f samples %d\n",
+		name, bucketLowerBound, bucketUpperBound, samples)
+}
+
+func (r *printStatsReporter) ReportHistogramDurationSamples(
+	name string,
+	tags map[string]string,
+	buckets tally.Buckets,
```
robskillington: Agreed, sure thing.

example/main.go
```diff
@@ -45,6 +45,30 @@ func (r *printStatsReporter) ReportTimer(name string, tags map[string]string, in
...
+	bucketLowerBound,
+	bucketUpperBound time.Duration,
+	samples int64,
+) {
+	fmt.Printf("histogram %s bucket lower %s upper %s samples %d\n",
+		name, bucketLowerBound.String(), bucketUpperBound.String(), samples)
```
robskillington: I don't mind either way, explicit or implicit.  Will make implicit.

histogram.go
```diff
@@ -0,0 +1,254 @@
...
+
+import (
+	"fmt"
+	"math"
+	"sort"
+	"time"
+)
+
+// Defaults is a type that can represent a constant.
+type Defaults int
```
robskillington: It's a little bit more explicit, however I'm not against making this nil and not differentiating between the two cases.  Will update this to `const DefaultBuckets Buckets = nil`.

histogram.go
```diff
@@ -0,0 +1,254 @@
...
+func (v Values) Less(i, j int) bool {
+	return v[i] < v[j]
+}
+
+func (v Values) String() string {
+	values := make([]string, len(v))
+	for i := range values {
+		values[i] = fmt.Sprintf("%f", v[i])
+	}
+	return fmt.Sprintf("%v", values)
```
robskillington: Sure thing.

histogram.go
```diff
@@ -0,0 +1,254 @@
...
+}
+
+// AsValues implements Buckets.
+func (v Values) AsValues() []float64 {
+	sort.Sort(v) // Always sort first
+	return []float64(v)
+}
+
+// AsDurations implements Buckets.
```
robskillington: Good call, will do.

histogram.go
```diff
@@ -0,0 +1,254 @@
...
+
+// BucketPairs creates a set of bucket pairs from a set
+// of buckets describing the lower and upper bounds for
+// each derived bucket.
+func BucketPairs(buckets Buckets) []BucketPair {
+	var (
+		asValueBuckets    = buckets.AsValues()
+		asDurationBuckets = buckets.AsDurations()
+	)
+	var pairs []BucketPair
```
robskillington: It's not necessarily the same length, in fact a lot of the time there's at least one added pair to add the range `[-infinity, first_bucket)` and `(last_bucket, infinity]`.

histogram.go
```diff
@@ -0,0 +1,254 @@
...
+}
+
+// LinearValueBuckets creates a set of linear value buckets.
+func LinearValueBuckets(start, width float64, count int) Values {
+	if count <= 0 {
+		panic("count needs to be > 0")
+	}
+	buckets := make([]float64, count)
+	for i := range buckets {
+		buckets[i] = start + (float64(i) * width)
```
robskillington: The value of previous upper bounds is derived from the buckets, similar to other open source client libraries.  I believe using such small values for measurement of float64's is outside the problem domain of this implementation of histograms.

For precise time measurements using histograms using the `DurationBuckets` entirely avoids this class of problems as its using `time.Duration` which are int64 nanoseconds.

histogram.go
```diff
@@ -0,0 +1,254 @@
...
+	curr := start
+	for i := range buckets {
+		buckets[i] = curr
+		curr *= factor
+	}
+	return Values(buckets)
+}
+
+// ExponentialDurationBuckets creates a set of exponential duration buckets.
+func ExponentialDurationBuckets(start time.Duration, factor float64, count int) Durations {
```
robskillington: So it won't actually allocate each time you call `scope.Histogram`.  The interface type `tally.Buckets` is allocated on the stack in all locations during histogram creation.  The histogram field `specification tally.Buckets` does not take a reference to the stack allocated interface passed in, sets the interface type with the `tab` and `data fields by value:
https://golang.org/src/runtime/runtime2.go?h=iface#L143

Take the following example:
```go
package main

import "testing"

var existingFooBar = &fooBar{}

type Foo interface {
	Foo()
}

type FooBar interface {
	Foo()
	Bar()
}

type fooBar struct {
	x int
}

func (*fooBar) Foo() {}
func (*fooBar) Bar() {}

type thing struct {
	foo Foo
}

func newThing(foo Foo) thing {
	return thing{foo: foo}
}

func BenchmarkIfaceAllocs(b *testing.B) {
	ref := existingFooBar
	for i := 0; i < b.N; i++ {
		x := newThing(ref)
		if x.foo != ref {
			b.Fail()
		}
//		ref = &fooBar{x: i}
	}
}
```

When run with `ref = &fooBar{x: i}` _not_ commented out it causes one allocation per op:
```
$ go test -bench Iface -v -benchmem
BenchmarkIfaceAllocs-4          100000000               17.6 ns/op             8 B/op            1 allocs/op
PASS
```

When run with `ref = &fooBar{x: i}` commented out it cases zero allocations per op:
```
$ go test -bench Iface -v -benchmem
BenchmarkIfaceAllocs-4          2000000000               1.28 ns/op            0 B/op            0 allocs/op
PASS
```

histogram.go
```diff
@@ -0,0 +1,254 @@
...
+	if count <= 0 {
+		panic("count needs to be > 0")
+	}
+	if start <= 0 {
+		panic("start needs to be > 0")
+	}
+	if factor <= 1 {
+		panic("factor needs to be > 1")
+	}
```
robskillington: I'm not against this, sure thing.

m3/reporter_test.go
robskillington: This was a carry over, I'll update this to simply `Equal`.

m3/reporter_test.go
robskillington: I was thinking about that, sure thing will DRY this up.

multi/reporter.go
robskillington: Good catch.  Yes this module needs coverage for the new histogram features.  I should've mentioned this was lacking and I was still working on it.

scope.go
robskillington: Of course, np.

reporter.go
robskillington: They can be inferred however during emission time to a non-cached reporter it is not very performant to walk the list of buckets and find the lower bound then format that into a stat name.

Passing these values makes it a relatively fast and straight forward implementation for a reporter like `statsd`:
```go
func (r *cactusStatsReporter) ReportHistogramValueSamples(
	name string,
	tags map[string]string,
	buckets tally.Buckets,
	bucketLowerBound,
	bucketUpperBound float64,
	samples int64,
) {
	r.statter.Inc(
		fmt.Sprintf("%s.%s-%s", name,
			r.valueBucketString(bucketLowerBound),
			r.valueBucketString(bucketUpperBound)),
		samples, r.sampleRate)
}

func (r *cactusStatsReporter) ReportHistogramDurationSamples(
	name string,
	tags map[string]string,
	buckets tally.Buckets,
	bucketLowerBound,
	bucketUpperBound time.Duration,
	samples int64,
) {
	r.statter.Inc(
		fmt.Sprintf("%s.%s-%s", name,
			r.durationBucketString(bucketLowerBound),
			r.durationBucketString(bucketUpperBound)),
		samples, r.sampleRate)
}
```

robskillington:
Re queries @akshayjshah :

> 1. Can you add some benchmarks for duration histograms? What sort of CPU and allocation cost should I expect for scope.Histogram("latency", tally.DefaultBuckets), and what should I expect for repeated calls to RecordDuration on that histogram (assuming a caching reporter, of course).

I can add these yes.  For a default call to `scope.Histogram("new-latency", tally.DefaultBuckets)` there should be a few for sure, for an existing histogram however that just returns an existing histogram by name it should be zero allocations.

Since histogram allocation should occur at startup time and is cached I wouldn't be too much concerned about the initial allocation.  It will be one of the more minimal allocation costs caused for a component inside your program that undergoes initialization, again, one time.

```
$ go test -bench BenchmarkHistogramExisting -benchmem
BenchmarkHistogramExisting-4      50000000                0 allocs/op
PASS
ok      github.com/uber-go/tally        1.746s
$ go test -bench BenchmarkHistogramAllocation -benchmem
BenchmarkHistogramAllocation-4            200000              44 allocs/op
PASS
ok      github.com/uber-go/tally        1.800s
```

Here is the M3 cached reporter benchmark exhibiting zero allocations at emission time for `RecordDuration`:

```
$ go test -bench BenchmarkScopeReportHistogram -benchmem
BenchmarkScopeReportHistogram-4         50000000                0 allocs/op
PASS
ok      github.com/uber-go/tally/m3     1.969s
```

> How attached are you to histograms that track floating-point values? If you make all histograms track int64s, you may get a much simpler implementation - you can just treat durations as an integer number of nanoseconds.

I would like to fully support floating point histograms as there are use cases people have already expressed interest in for such a type of histogram.  Also it allows one to use the full extent of the types supported by the reporters.  Otherwise they would be missing the ability using tally to report floats as histogram even though the underlying reporters support it.

> If you'd like a detailed code review, this is a lot for a relative Tally newcomer. It'd be easier to review (for me, at least) as a series of stacked diffs - interfaces, then statsd implementation, then m3, etc.

Of course, I just wanted to make sure all your concerns are met re: making histograms available to all users of Tally in a clean and pragmatic approach.  Xi, Jake and others can help with the full review.

m3/reporter.go
```diff
@@ -439,6 +537,62 @@ func (c cachedMetric) ReportTimer(interval time.Duration) {
...
+
+func (h cachedHistogram) ValueBucket(
+	bucketLowerBound, bucketUpperBound float64,
+) tally.CachedHistogramBucket {
+	for _, b := range h.cachedValueBuckets {
+		if b.valueUpperBound >= bucketUpperBound {
+			return b.metric
+		}
+	}
+	return noopMetric{}
```
xichen2020: Might make sense to return error here, can also check if lower bound is smaller than upper bound etc

prometheus/reporter.go
```diff
@@ -185,11 +186,42 @@ func (m *cachedMetric) reportTimerSummary(interval time.Duration) {
...
+}
+
+type cachedHistogramBucket struct {
+	metric     *cachedMetric
+	upperBound float64
+}
+
+func (b cachedHistogramBucket) ReportSamples(value int64) {
+	for i := int64(0); i < value; i++ {
+		b.metric.histogram.Observe(b.upperBound)
```
xichen2020: Is upper bound exclusive or inclusive?

scope.go
```diff
@@ -223,14 +246,13 @@ func (s *scope) Counter(name string) Counter {
 		s.cm.Lock()
 		val, ok = s.counters[name]
 		if !ok {
+			var cachedCounter CachedCount
```
xichen2020: Curious, why is the type not called `CachedCounter`?

scope.go
```diff
@@ -269,25 +290,50 @@ func (s *scope) Timer(name string) Timer {
...
 
+func (s *scope) Histogram(name string, b Buckets) Histogram {
+	if b == nil {
+		b = s.defaultBuckets
+	}
+
+	s.hm.RLock()
+	val, ok := s.histograms[name]
+	s.hm.RUnlock()
+	if !ok {
```
xichen2020: nit: could reduce nesting here by checking `if ok` first

stats.go
```diff
@@ -193,27 +207,233 @@ func (t *timer) snapshot() []time.Duration {
...
+
+	h := &histogram{
+		htype:            htype,
+		name:             name,
+		tags:             tags,
+		reporter:         reporter,
+		specification:    buckets,
+		buckets:          make([]histogramBucket, 0, len(pairs)),
+		lookupByValue:    make([]float64, 0, len(pairs)),
+		lookupByDuration: make([]int, 0, len(pairs)),
```
xichen2020: Should this be `int64` to be consistent?

stats.go
```diff
@@ -193,27 +207,233 @@ func (t *timer) snapshot() []time.Duration {
...
+		case valueHistogramType:
+			h.buckets[i].cachedValueBucket.ReportSamples(samples)
+		case durationHistogramType:
+			h.buckets[i].cachedDurationBucket.ReportSamples(samples)
+		}
+	}
+}
+
+func (h *histogram) RecordValue(value float64) {
+	idx := sort.SearchFloat64s(h.lookupByValue, value)
```
xichen2020: Need a comment here documenting the lowest value in `lookupByValue` is the min float64 and the highest value in `lookupByValue` is max float64

stats.go
```diff
@@ -193,27 +207,233 @@ func (t *timer) snapshot() []time.Duration {
...
+	h.buckets[idx].samples.Inc(1)
+}
+
+func (h *histogram) RecordDuration(value time.Duration) {
+	idx := sort.SearchInts(h.lookupByDuration, int(value))
```
xichen2020: Same here, needs better comment

statsd/reporter.go
```diff
@@ -64,6 +85,60 @@ func (r *cactusStatsReporter) ReportTimer(name string, tags map[string]string, i
...
+			r.durationBucketString(bucketLowerBound),
+			r.durationBucketString(bucketUpperBound)),
+		samples, r.sampleRate)
+}
+
+func (r *cactusStatsReporter) valueBucketString(
+	upperBound float64,
+) string {
+	if upperBound == math.MaxFloat64 {
+		return "infinity"
```
xichen2020: Make them consts?

histogram.go
```diff
@@ -0,0 +1,259 @@
...
+
+// AsValues implements Buckets.
+func (v ValueBuckets) AsValues() []float64 {
+	return []float64(v)
+}
+
+// AsDurations implements Buckets and returns time.Duration
+// representations of the float64 values divided by time.Second.
+func (v ValueBuckets) AsDurations() []time.Duration {
+	values := make([]time.Duration, len(v))
```
robskillington: This is true however none of our other interfaces return an error on a counter inc or gauge update or timer record.  I can perhaps make it a no-op if you `RecordValue(...)` on a histogram that has duration buckets and vice-versa if you feel strongly?

histogram.go
```diff
@@ -0,0 +1,259 @@
...
+	values := make([]string, len(v))
+	for i := range values {
+		values[i] = v[i].String()
+	}
+	return fmt.Sprintf("%v", values)
+}
+
+// AsValues implements Buckets and returns float64
+// representations of the time.Duration values divided by time.Second.
+func (v DurationBuckets) AsValues() []float64 {
```
robskillington: As above.

histogram.go
```diff
@@ -0,0 +1,259 @@
...
+	return []time.Duration(v)
+}
+
+// BucketPairs creates a set of bucket pairs from a set
+// of buckets describing the lower and upper bounds for
+// each derived bucket.
+func BucketPairs(buckets Buckets) []BucketPair {
+	if buckets == nil || buckets.Len() < 1 {
+		return []BucketPair{
+			bucketPair{
```
robskillington: Sure thing.

histogram.go
```diff
@@ -0,0 +1,259 @@
...
+				-math.MaxFloat64, math.MaxFloat64,
```
robskillington: Sure thing, will update.

histogram.go
```diff
@@ -0,0 +1,259 @@
...
+}
+
+func (p bucketPair) UpperBoundDuration() time.Duration {
+	return p.upperBoundDuration
+}
+
+// LinearValueBuckets creates a set of linear value buckets.
+func LinearValueBuckets(start, width float64, count int) ValueBuckets {
+	if count <= 0 {
+		panic("count needs to be > 0")
```
robskillington: Yeah I'm going to return errors then add separate methods such as `MustMakeLinearValueBuckets(...)`, etc to return a single value for fluid API use.

m3/reporter.go
```diff
@@ -58,8 +59,17 @@ const (
...
 	// DefaultMaxPacketSize is the default M3 reporter max packet size.
 	DefaultMaxPacketSize = int32(1440)
-
-	emitMetricBatchOverhead = 19
+	// DefaultHistogramBucketTagPrecision is the default
+	// precision to use when formatting the metric tag
+	// with the histogram bucket bound values.
+	DefaultHistogramBucketTagPrecision = uint(6)
+	// HistogramBucketIDTagName is the histogram bucket ID tag name
+	HistogramBucketIDTagName = "bucketid"
```
robskillington: Sure thing.

m3/reporter.go
```diff
@@ -207,6 +221,7 @@ func NewReporter(opts Options) (Reporter, error) {
 		commonTags:   tags,
 		freeBytes:    freeBytes,
 		resourcePool: resourcePool,
+		bucketValFmt: "%." + strconv.Itoa(int(opts.HistogramBucketTagPrecision)) + "f",
```
robskillington: This is simply on allocation time.  Do you feel strongly about this?

m3/reporter.go
robskillington: This is tally itself breaking its own contracts.  I can panic if you'd like.  Otherwise the interface starts to look very not great.

prometheus/reporter.go
robskillington: Inclusive - as per other histogram implementations, like Prometheus, et al.

scope.go
robskillington: I did not create these interfaces, unsure.

stats.go
robskillington: It needs to be `[]int` so we can perform a `sort.SearchInts(...)` efficiently.

statsd/reporter.go
robskillington: For code readability I'm inclined to leave them here.  They will be consts for essentially after compilation.

stats.go
robskillington: Sure thing.

stats.go
robskillington: Sure thing.

PR #30, opened 2017-02-12
[Support Messagepack Encoding - PART 1] Refactor M3 reporter to support multiple encodings

jeromefroe:
Recent work has shown that encoding metrics using Messagepack may be more efficient than using Thrift. Consequently, I'd like to introduce a new M3 reporter which sends metrics over the wire using Messagepack instead of Thrift. This first PR refactors the current M3 reporter code to make it easier to add a Messagepack reporter. Subsequent PR's will introduce the actual Messagepack reporter.


jeromefroe:
Talked with Rob a while back and we decided it would be better to move the m3 reporters into a repo under m3db, so I'm going to close this PR in favor of doing that later on.

PR #32, opened 2017-02-19
Add zero allocation Stopwatch for both timer and histograms

robskillington:
It's still a little slower than the default `start := time.Now(); timer.Record(time.Since(start))` but it doesn't cause any allocations any longer.

```
$ go test -v -run none -bench BenchmarkTimerStopwatch -benchmem
BenchmarkTimerStopwatch-4       30000000                51.3 ns/op             0 B/op          0 allocs/op
PASS
ok      github.com/uber-go/tally        1.726s

$ go test -v -run none -bench BenchmarkTimerReport -benchmem
BenchmarkTimerReport-4          50000000                36.5 ns/op             0 B/op          0 allocs/op
PASS
ok      github.com/uber-go/tally        2.005s
```

cc @xichen2020 @akshayjshah @Raynos

### DIFF
--- a/stats.go
+++ b/stats.go
@@ -181,11 +181,11 @@ func (t *timer) Record(interval time.Duration) {
 }
 
 func (t *timer) Start() Stopwatch {
-	return Stopwatch{Start: globalClock.Now(), Recorder: t}
+	return NewStopwatch(globalClock.Now(), t)
 }
 
-func (t *timer) RecordStopwatch(sw Stopwatch) {
-	d := globalClock.Now().Sub(sw.Start)
+func (t *timer) RecordStopwatch(stopwatchStart time.Time) {
+	d := globalClock.Now().Sub(stopwatchStart)
 	t.Record(d)
 }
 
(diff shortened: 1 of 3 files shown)


types.go
```diff
@@ -95,10 +95,23 @@ type Histogram interface {
...
 
-// Stopwatch is a helper for simpler tracking of elapsed time.
-type Stopwatch interface {
-	// Stop records the difference between the current clock and start time.
-	Stop()
+// Stopwatch is a helper for simpler tracking of elapsed time, use the
+// Stop() method to report time elapsed since its created back to the
+// timer or histogram.
+type Stopwatch struct {
+	Start    time.Time
```
xichen2020: Can we make the fields unexported to encapsulate implementation details?

robskillington: I was originally going to do that but I thought this may be more useful as the type can be used outside the package for return from a subpackage type.

stats.go
```diff
@@ -375,7 +367,12 @@ func (h *histogram) RecordDuration(value time.Duration) {
 }
 
 func (h *histogram) Start() Stopwatch {
-	return histogramStopwatch{start: globalClock.Now(), histogram: h}
+	return NewStopwatch(globalClock.Now(), h)
+}
+
+func (h *histogram) RecordStopwatch(stopwatchStart time.Time) {
```
prashantv: This name seems specific to the use case of `Stopwatch` below, should we consider a more generic name like `RecordSince(time.Time)` which is similar to [time.Since](https://golang.org/pkg/time/#Since) in the standard library

PR #33, opened 2017-02-19
Fix possible flush after close semantics for scope

robskillington:
There was the possibility a reporter might have `Flush` or any of the other report methods called on it after it had been closed itself.  This change fixes that possibility by holding a status lock during each report loop run.

cc @xichen2020

### DIFF
--- a/scope.go
+++ b/scope.go
@@ -66,7 +66,7 @@ type scope struct {
 	defaultBuckets Buckets
 
 	registry *scopeRegistry
-	status   *scopeStatus
+	status   scopeStatus
 
 	cm sync.RWMutex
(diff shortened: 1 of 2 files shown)


scope.go
```diff
@@ -66,7 +66,7 @@ type scope struct {
 	defaultBuckets Buckets
 
 	registry *scopeRegistry
-	quit     chan struct{}
+	status   *scopeStatus
```
xichen2020: Can just be `scopeStatus` to reduce level of indirection?

robskillington: Since it does not need to be passed to child scopes I can do this, I'll update it.

Issue #34, opened 2017-03-07
Snapshot cannot distinguish metrics with the same fully qualified name but different tags


jeromefroe:
Presently, `Snapshot` cannot distinguish between metrics that have the same fully qualified name but different tags. Consequently, when writing unit tests, one cannot test the values for metrics with the same name.

For example, consider the following counters on a given scope:

```
fooRequests := scope.Tagged(map[string]string{"endpoint": "/foo"}.Counter("requests")
barRequests := scope.Tagged(map[string]string{"endpoint": "/bar"}.Counter("requests")
```

`Snapshot` stores metrics in a map using only their fully qualified name. Since these metrics have the same fully qualified name, "requests" in this case, when the second metric is inserted into the map created by `Snapshot` it will overwrite the first metric:

```
	for _, ss := range s.registry.subscopes {
		...
		for key, c := range ss.counters {
			name := ss.fullyQualifiedName(key)
			snap.counters[name] = &counterSnapshot{
				name:  name,
				tags:  tags,
				value: c.snapshot(),
			}
		}
		..
      }
```

jeromefroe:
Closing this since #35 was merged.

PR #35, opened 2017-03-08
Use a unique ID for metrics returned from Snapshot

jeromefroe:
This PR addresses #34 by introducing a new function `uniqueID` which will create a unique ID for a metric using it fully qualified name and tags. This ID is then used as the key for the maps created `Snapshot` so that metrics with the same fully qualified names but different tags will not overwrite one another.

### DIFF
--- a/scope.go
+++ b/scope.go
@@ -424,6 +424,7 @@ func (s *scope) Capabilities() Capabilities {
 }
 
 func (s *scope) Snapshot() Snapshot {
+	var buf bytes.Buffer
 	snap := newSnapshot()
 
 	s.registry.RLock()
@@ -437,7 +438,7 @@ func (s *scope) Snapshot() Snapshot {
 		ss.cm.RLock()
 		for key, c := range ss.counters {
 			name := ss.fullyQualifiedName(key)
-			id := ss.uniqueID(name, tags)
+			id := ss.uniqueID(name, tags, buf)
 			snap.counters[id] = &counterSnapshot{
 				name:  name,
 				tags:  tags,
@@ -448,7 +449,7 @@ func (s *scope) Snapshot() Snapshot {
 		ss.gm.RLock()
 		for key, g := range ss.gauges {
 			name := ss.fullyQualifiedName(key)
-			id := ss.uniqueID(name, tags)
+			id := ss.uniqueID(name, tags, buf)
 			snap.gauges[id] = &gaugeSnapshot{
 				name:  name,
 				tags:  tags,
@@ -459,7 +460,7 @@ func (s *scope) Snapshot() Snapshot {
 		ss.tm.RLock()
 		for key, t := range ss.timers {
 			name := ss.fullyQualifiedName(key)
-			id := ss.uniqueID(name, tags)
+			id := ss.uniqueID(name, tags, buf)
 			snap.timers[id] = &timerSnapshot{
 				name:   name,
 				tags:   tags,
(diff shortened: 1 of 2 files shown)


scope.go
```diff
@@ -429,7 +437,8 @@ func (s *scope) Snapshot() Snapshot {
 		ss.cm.RLock()
 		for key, c := range ss.counters {
 			name := ss.fullyQualifiedName(key)
-			snap.counters[name] = &counterSnapshot{
+			id := ss.uniqueID(key, tags)
```
xichen2020: Perhaps just pass `name` in here so you don't need to call `fullQualifiedName` another time in `uniqueID`?

scope.go
```diff
@@ -482,6 +493,43 @@ func (s *scope) fullyQualifiedName(name string) string {
...
 }
 
+func (s *scope) uniqueID(name string, tags map[string]string) string {
+	name = s.fullyQualifiedName(name)
+	l := len(name)
+	for k, v := range tags {
+		l += len(k) + len(v)
+	}
+
+	// size the buffer to include separator bytes
```
xichen2020: `Size of ...`

scope.go
```diff
@@ -482,6 +493,43 @@ func (s *scope) fullyQualifiedName(name string) string {
...
+	buf := bytes.NewBuffer(make([]byte, 0, l+2*len(tags)+1))
```
xichen2020: Can we reuse this buffer across multiple calls of `uniqueID`?

xichen2020: Also should it just be `l + 2 * len(tags)`?

scope.go
```diff
@@ -482,6 +493,43 @@ func (s *scope) fullyQualifiedName(name string) string {
...
+	buf.WriteString(name)
+
+	if len(tags) == 0 {
```
xichen2020: We should move this check before the buffer creation so we don't need to create the buffer and return name immediately if there are no tags

scope.go
```diff
@@ -34,6 +36,12 @@ var (
 	NoopScope, _ = NewRootScope(ScopeOptions{Reporter: NullStatsReporter}, 0)
 	// DefaultSeparator is the default separator used to join nested scopes
 	DefaultSeparator = "."
+	// ComponentSeparator is the separator used to separate a metric's name and it's tags in its ID
+	ComponentSeparator = "+"
```
xichen2020: Wonder if we should make these unexported, they seem to be internal implementation details to the scope for computing unique IDs

scope.go
```diff
@@ -482,6 +493,43 @@ func (s *scope) fullyQualifiedName(name string) string {
...
+	buf := bytes.NewBuffer(make([]byte, 0, l+2*len(tags)+1))
+	buf.WriteString(name)
+
+	if len(tags) == 0 {
+		return buf.String()
+	}
+
+	buf.WriteString(ComponentSeparator)
+
+	// sort tags in alphabetical order so the ID is consistent
```
xichen2020: Conventionally the first letter in a comment should be upper case

scope.go
jeromefroe: Done.

scope.go
jeromefroe: Done.

scope.go
jeromefroe: Done.

scope.go
jeromefroe: Done.

scope.go
jeromefroe: Done.

scope.go
jeromefroe: We could add a `Buffer` as a field of a `scope` and reuse it in calls to `Snapshot`. We would have to add a lock for it though to guard against concurrent calls to `uniqueID` from concurrent calls to `Snapshot`. I kind of like that approach.

jeromefroe: Done.

xichen2020: Alternatively we can just create a buffer at the beginning of the `Snapshot` function, and pass the buffer to `uniqueID` function to reuse it across calls. This way we don't need to lock it and doesn't need to store it inside the scope either (which affects the space usage of a real scope)

jeromefroe: Good idea, done.

scope.go
```diff
@@ -449,7 +460,8 @@ func (s *scope) Snapshot() Snapshot {
 		ss.tm.RLock()
 		for key, t := range ss.timers {
 			name := ss.fullyQualifiedName(key)
-			snap.timers[name] = &timerSnapshot{
+			id := ss.uniqueID(name, tags, buf)
```
robskillington: Hey instead of adding a new method `uniqueID` its probably best to use the already existing method `func KeyForPrefixedStringMap(prefix string, stringMap map[string]string) string` which resides in `key_gen.go`.

jeromefroe: Good point, didn't realize we had such a function already.

xichen2020: lol fair enough

robskillington:
We'll have to make sure we release this as 3.0 as it'll break compatibility somewhat.

PR #39, opened 2017-03-20
Add HTTP endpoint to view metric snapshots

jeromefroe:
This PR adds the ability to get a HTTP handler for a Scope which users can use to add an endpoint to a HTTP server to view metric snapshots when debugging. It's not finished yet but I thought it might be helpful to put it up now to solicit feedback. I will update the PR when I finish the changes I want to make.


jeromefroe:
Closing in favor of 52.

PR #40, opened 2017-03-21
Fix licence link by inserting newline

sectioneight:
Because of the inline HTML, GitHub wasn't parsing the markdown correctly.

Inserting a newline fixes this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -194,6 +194,7 @@ BenchmarkTimerReport-8              	300000000	         5.69 ns/op
 ```
 
 <hr>
+
 Released under the [MIT License](LICENSE).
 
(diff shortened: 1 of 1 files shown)


sectioneight:
Before:

screen shot 2017-03-20 at 10 45 17 pm


After:

screen shot 2017-03-20 at 10 45 38 pm

sectioneight:
Clickable preview here: https://github.com/uber-go/tally/blob/a25bbcc43c482b8e3d621407d969e052a15e3c62/README.md#performance

robskillington:
Ah brilliant, thanks for this @sectioneight

PR #44, opened 2017-05-12
[scope] Report buffered metrics on Close()

schallert:
Before flushing the underlying reporter we need to ensure that all
metrics buffered in the scope are reported as well.


scope.go
```diff
@@ -478,9 +478,12 @@ func (s *scope) Snapshot() Snapshot {
 }
 
 func (s *scope) Close() error {
+	// prevent any more report loops via tick before forcing report
+	close(s.status.quit)
```
martin-mao: We should probably add a check to ensure that we don't panic due to closing chan twice by calling close twice.

scope.go
```diff
@@ -478,10 +474,12 @@ func (s *scope) Snapshot() Snapshot {
...
-	close(s.status.quit)
-	s.status.Unlock()
+	if atomic.CompareAndSwapInt32(&s.status.closed, 0, 1) {
+		// prevent any more report loops via tick before forcing report
+		close(s.status.quit)
+		s.reportLoopRun()
+		atomic.StoreInt32(&s.status.closed, 2)
+	}
 
 	if closer, ok := s.baseReporter.(io.Closer); ok {
```
martin-mao: If we call close and it's already closed, do we want to close the baseReporter? It's already been closed once no?

scope.go
```diff
@@ -478,10 +474,12 @@ func (s *scope) Snapshot() Snapshot {
 }
 
 func (s *scope) Close() error {
-	s.status.Lock()
-	s.status.closed = true
-	close(s.status.quit)
-	s.status.Unlock()
+	if atomic.CompareAndSwapInt32(&s.status.closed, 0, 1) {
+		// prevent any more report loops via tick before forcing report
+		close(s.status.quit)
```
martin-mao: Do we have a potential race here if reportLoopRun is triggered after line 477 with the tick, then when you come into here, there's actually 2 reportLoopRun functions running and you only block on one of them completing (ie the other one can still be running when you return from this function).

This may be especially true if you close the baseReporter right afterwards, ie some metrics are not yet flushed.

scope.go
```diff
@@ -239,14 +241,13 @@ func (s *scope) reportLoop(interval time.Duration) {
...
-	// and flush after a close
-	s.status.RLock()
-	if s.status.closed {
-		s.status.RUnlock()
+	// don't flush after scope fully closed
+	if atomic.LoadInt32(&s.status.closed) == 2 {
 		return
 	}
 
+	s.status.flushes.Add(1)
```
martin-mao: This may have to be done outside of reportLoopRun ie at the callers otherwise there's a race of wait being called when someone calls this and is at line 248.

Issue #45, opened 2017-06-21
Investigate panic in M3 reporter 


robskillington:
```go
panic in handler: send on closed channel
code.uber.internal/recognition/tips/vendor/go.uber.org/fx/modules/yarpc.panicRecovery
	code.uber.internal/recognition/tips/vendor/go.uber.org/fx/modules/yarpc/middleware.go:143
runtime.call32
	/usr/lib/go-1.8/src/runtime/asm_amd64.s:514
runtime.gopanic
	/usr/lib/go-1.8/src/runtime/panic.go:489
runtime.chansend
	/usr/lib/go-1.8/src/runtime/chan.go:180
runtime.selectnbsend
	/usr/lib/go-1.8/src/runtime/chan.go:604
code.uber.internal/recognition/tips/vendor/github.com/uber-go/tally/m3.(*reporter).reportCopyMetric
	code.uber.internal/recognition/tips/vendor/github.com/uber-go/tally/m3/reporter.go:431
code.uber.internal/recognition/tips/vendor/github.com/uber-go/tally/m3.cachedMetric.ReportTimer
	code.uber.internal/recognition/tips/vendor/github.com/uber-go/tally/m3/reporter.go:543
code.uber.internal/recognition/tips/vendor/github.com/uber-go/tally/m3.(*cachedMetric).ReportTimer
	<autogenerated>:11
code.uber.internal/recognition/tips/vendor/github.com/uber-go/tally.(*timer).Record
	code.uber.internal/recognition/tips/vendor/github.com/uber-go/tally/stats.go:177
code.uber.internal/recognition/tips/vendor/github.com/uber-go/tally.(*timer).RecordStopwatch
	code.uber.internal/recognition/tips/vendor/github.com/uber-go/tally/stats.go:189
code.uber.internal/recognition/tips/vendor/github.com/uber-go/tally.Stopwatch.Stop
	code.uber.internal/recognition/tips/vendor/github.com/uber-go/tally/types.go:114
code.uber.internal/recognition/tips/vendor/code.uber.internal/recognition/glue.git/decorators.Metrics.func1.1
	code.uber.internal/recognition/tips/vendor/code.uber.internal/recognition/glue.git/decorators/metrics.go:29
code.uber.internal/recognition/tips/vendor/code.uber.internal/recognition/glue.git/decorators.YARPCScope.func1.1
	code.uber.internal/recognition/tips/vendor/code.uber.internal/recognition/glue.git/decorators/scope.go:36
[...]

PR #46, opened 2017-06-22
Avoid send on channel when emitting values after closing an M3 reporter

robskillington:
Addresses https://github.com/uber-go/tally/issues/45

cc @mcgourty @xichen2020 @martin-mao

### DIFF
--- a/m3/reporter.go
+++ b/m3/reporter.go
@@ -427,6 +427,14 @@ func (r *reporter) reportCopyMetric(
 		copy.MetricValue.Timer = t
 	}
 
+	// NB(r): This is to avoid sending on a closed channel,
+	// it's faster to actually defer/recover than acquire
+	// a read lock here to ensure we aren't closed: benchmarked
+	// this with BenchmarkTimer in reporter_benchmark_test.go
+	defer func() {
+		recover()
+	}()
+
 	select {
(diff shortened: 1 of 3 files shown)


mcgourty:
thanks for doing this!

m3/reporter.go
```diff
@@ -427,6 +427,14 @@ func (r *reporter) reportCopyMetric(
 		copy.MetricValue.Timer = t
 	}
 
+	// NB(r): This is to avoid sending on a closed channel,
+	// it's faster to actually defer/recover than acquire
+	// a read lock here to ensure we aren't closed: benchmarked
+	// this with BenchmarkTimer in reporter_benchmark_test.go
+	defer func() {
```
xichen2020: This is a somewhat unconventional way to avoid sending to closed channels. When the process recovers, it's hard to know whether it is in a position to swallow the panic and proceed as if nothing had happened. Is the overhead of acquiring read locks too prohibitive to be practical?

robskillington: It's roughly 350ns vs 630ns.  That's a pretty big price considering that timers are very high QPS.

xichen2020: hm okay, I guess when the reporter is closed, the process is probably shutting down anyway.

PR #50, opened 2017-07-25
Ability to sanitise metric identifiers

prateek:
Added the ability for users to sanitise metrics. 

Benchmark results:

#### Master
```
❯ go test -bench=NewMetric ./m3
BenchmarkNewMetric-4   	10000000	       183 ns/op
PASS
ok  	github.com/uber-go/tally/m3	2.364s
```

#### PR
```
❯ go test -bench=NewMetric ./m3
BenchmarkNewMetric-4                   	10000000	       194 ns/op
BenchmarkNewMetricWithSanitization-4   	 5000000	       413 ns/op
PASS
ok  	github.com/uber-go/tally/m3	4.816s
```

i.e. 6% slowdown for `newMetric` without sanitisation, and sanitised metrics are 125% slower than non-sanitised metrics.


m3/config.go
```diff
@@ -45,6 +45,9 @@ type Configuration struct {
 
 	// IncludeHost is whether or not to include host tag.
 	IncludeHost bool `yaml:"includeHost"`
+
+	// SanitizeMetrics is whether or not metric identifiers (name, tags) are sanitized.
+	SanitizeMetrics bool `yaml:"sanitizeMetrics"`
```
jeromefroe: Instead of making this a `bool` perhaps it would be preferable to expose `sanitizeFn` as a public type and make this a `SanitizeFn`. The default sanitizer could be a noop function, but we could also provide a few different sanitizer functions that correspond to different reporters. For example, a statsd reporter would only replace the characters that are used as delimiters in the statsd protocol and the current sanitizer could be an alphanumeric sanitizer.

m3/util.go
```diff
@@ -0,0 +1,61 @@
...
+	defaultReplacementCharacter = '_'
+)
+
+var (
+	validRangeCharacters  = [][2]int{{int('a'), int('z')}, {int('A'), int('Z')}, {int('0'), int('9')}}
+	validSingleCharacters = []rune{'-', '_'}
+)
+
+func sanitizeMetricIdentifiers(value string) string {
+	buf := bytes.NewBuffer(make([]byte, 0, len(value)))
```
robskillington: To avoid allocations in the general case we can lazily construct this buffering similar to how some parts of the Go stdlib does replacements:
```go
var buf *bytes.Buffer
for i, ch := range value {
  // ... use existing code to calculate validCurr ...

  if validCurr {
    if buf == nil {
      // Still don't need to deviate from string
      continue
    }
    buf.WriteRune(ch)
    continue
  }

  if buf == nil {
    // Definitely have deviated, need to initialize buf and backfill
    buf = bytes.NewBuffer(make([]byte, 0, len(value)))
    if i > 0 {
      buf.WriteString(value[:i-1])
    }
  }
  buf.WriteRune(defaultReplacementCharacter)
}

func lazyInitBuf(buf *bytes.Buffer) 
```

m3/config.go
robskillington: I'm supportive of this as well, I think by default just having a `nil` value for `SanitizeFn` is best and we can provide a `SanitizeAlphanumericFn` and a `SanitizeStatsdFn` as out of the box sanitizers people can set the `SanitizeFn` in the Scope options too.

prateek: Sure, wrote up a first pass of what this could look like. Take a look.

m3/util.go
prateek: The allocation trick helped a lot, perf impact is almost negligible because of it.

prateek: Re: `validLast`, consider the case when underscore is a valid character, the sanitised output for `abc_:` v `abc_` is identical. That feels weird. 

IMO the behaviour is easier to explain if we only do a strict replace.

prateek:
New benchmarks: 

#### Master

```
❯ go test -bench=Name
BenchmarkNameGeneration-4           	  5000000	       320 ns/op
BenchmarkNameGenerationTagged-4     	  5000000	       319 ns/op
BenchmarkNameGenerationNoPrefix-4   	500000000	       3.58 ns/op
PASS
ok  	github.com/uber-go/tally	6.149s
```

#### PR
```
❯ go test -bench=Name
BenchmarkNameGeneration-4                    	  5000000	         324 ns/op
BenchmarkSanitisedNameGeneration-4            	  5000000	         325 ns/op
BenchmarkNameGenerationTagged-4              	  5000000	         332 ns/op
BenchmarkSanitisedNameGenerationTagged-4     	  5000000	         324 ns/op
BenchmarkNameGenerationNoPrefix-4              	500000000	         3.04 ns/op
BenchmarkSanitisedNameGenerationNoPrefix-4   	500000000	         3.06 ns/op
PASS
ok  	github.com/uber-go/tally	11.624s
```

sanitise.go
```diff
@@ -0,0 +1,192 @@
...
+}
+
+func (c *ValidCharacters) sanitiseFn(repChar rune) SanitiseFn {
+	return func(value string) string {
+		var buf *bytes.Buffer
+		for idx, ch := range value {
+			// first check if the provided character is valid
+			validCurr := false
+			for i := 0; !validCurr && i < len(c.Ranges); i++ {
+				for j := c.Ranges[i][0]; j <= c.Ranges[i][1]; j++ {
```
jeromefroe: Any reason we don't check the bounds directly here? Was thinking something along the lines of

```
if ch >= c.Ranges[i][0] && ch <= c.Ranges[i][1] {
    validCurr = true
}
```

jeromefroe:
Not sure I can approve the Britishification of sanitize to sanitise :)

sanitise.go
```diff
@@ -0,0 +1,192 @@
...
+		'.',
+		'-',
+		'_'}
+
+	// AlphanumericSanitiserOpts is the options to create a sanitiser which uses
+	// the alphanumeric SanitiseFn.
+	AlphanumericSanitiserOpts = SanitiseOptions{
+		NameCharacters: ValidCharacters{
+			Ranges:     AlphanumericRange,
+			Characters: UnderscoreDashCharacters,
```
xichen2020: Should this be `UnderscoreDashDotCharacters`?

sanitise.go
```diff
@@ -0,0 +1,192 @@
...
+		valueFn: opts.ValueCharacters.sanitiseFn(opts.ReplacementCharacter),
+	}
+}
+
+// NoOpSanitiseFn returns the input un-touched.
+func NoOpSanitiseFn(v string) string { return v }
+
+// NewNoOpSanitiser returns a sanitiser which returns all inputs un-touched.
+func NewNoOpSanitiser() Sanitiser {
+	return &santiser{
```
xichen2020: nit: no need for the pointer here.

sanitise.go
prateek: No reason at all, added in the next versions 👍

sanitise.go
prateek: sure thing

sanitise.go
prateek: Does that apply to both statsd and `m3`?

xichen2020: Does statsd have a concept of `name` though?

prateek: Dunno what makes sense for `statsd`. 

I made `AlphanumericSanitiserOpts` un-exported. We can let each reporter package export it's default sanitiser opts, similar to `m3/sanitise.go` in this PR.

scope_test.go
```diff
@@ -529,6 +575,36 @@ func TestTaggedSubScope(t *testing.T) {
...
+	tscope := root.Tagged(map[string]string{"service": "test.service"}).(*scope)
+
+	r.cg.Add(1)
+	tscope.Counter("beep").Inc(1)
+
+	s.report(r)
+	tscope.report(r)
+	r.cg.Wait()
+
+	assert.EqualValues(t, 1, r.counters["foo.beep"].val)
```
prateek: Wanted feedback on this behaviour – the implementation sanitises at the time of metric/scope construction. i.e. when we call `SubScope(_)`, or `Tagged()`, or `Counter(...)`. This means prefix concatenation can introduce chars invalid as defined by the sanitiser.

This example, we're specifying a prefix of `foo`, and permit `[a-zA-Z0-9-_]` as valid chars. When we create counter `foo`, we still end up with a counter named `foo.beep`. 

I prefer this behaviour to sanitising post the `fullyQualifiedName` call but I can see how it might be confusing to users. What do you think?

prateek: cc @robskillington @jeromefroe @xichen2020

sanitise.go
```diff
@@ -145,7 +147,7 @@ func (c *ValidCharacters) sanitiseFn(repChar rune) SanitiseFn {
 			// first check if the provided character is valid
 			validCurr := false
 			for i := 0; !validCurr && i < len(c.Ranges); i++ {
-				for j := c.Ranges[i][0]; j <= c.Ranges[i][1]; j++ {
+				for j := c.Ranges[i][0]; ch >= c.Ranges[i][0] && j <= c.Ranges[i][1]; j++ {
```
xichen2020: Can we do
```
if ch >= c.Ranges[i][0] && ch <= c.Ranges[i][1] {
  validCurr = true
  break
}
```
here?

scope.go
```diff
@@ -410,9 +410,9 @@ func (s *scope) subscope(prefix string, tags map[string]string) Scope {
 	subscope := &scope{
 		separator: s.separator,
 		prefix:    prefix,
-		// NB(r): Take a copy of the tags on creation
-		// so that it cannot be modified after set.
-		tags:           copyStringMap(tags),
+		// NB(prateek): don't need to copy the tags here,
+		// we assume the map provided is immutable.
+		tags:           tags,
```
xichen2020: Perhaps just rename `tags` to `immutableTags` then?

scope_test.go
xichen2020: Will both approaches lead to the same set of legal ids post sanitization? If so, I think it's probably okay as how the sanitization is implemented is an internal detail.

Out of curiosity, what's the benefit of sanitizing prefix and the name separately? I'm assuming that adds less overhead compared to sanitizing the fully qualified name?

scope.go
prateek: sure thing

scope_test.go
robskillington: Sanitizing the output of fullyQualifiedName seems more like what the user is asking for if they pass you a sanitizer, even if it's not "as powerful" per se.  i.e. no way to get around it if we do that.  

For instance if we do it on the output of fullyQualifiedName then for scopes using the Prometheus reporter we wouldn't have to set the default tag separator to something different, we would just have to set the sanitizer to UnderscoreDashSeparated and we'd be done - we get sanitization for our input and for the prefixing too.

robskillington:
Looking good, did respond to your prefixing comment though

scope_test.go
prateek: I'll buy that. Behaviour will be easier to explain if we sanitise everything too. Updated the code to do the sanitisation post `fullyQualifiedName`

scope.go
prateek: done

sanitise.go
prateek: yep, done

PR #51, opened 2017-07-31
Americanise sanitise

prateek:
- s/sanitise/sanitize 

/cc @robskillington @jeromefroe


m3/sanitise.go
```diff
@@ -25,8 +25,8 @@ import (
 )
 
 var (
```
robskillington: Need to rename this file too =[

prateek: Renamed :)

PR #54, opened 2017-08-30
Implement InstrumentedCall

proflayton:
go-common's metric interface includes this functionality. It would be great to include it in tally as well since it simplifies application code, we don't need to explicitly handle timers and error cases every time we want to include metrics

### DIFF
--- a/instrument.go
+++ b/instrument.go
@@ -36,10 +36,6 @@ func NewInstrumentedCall(scope Scope, name string) InstrumentedCall {
 	}
 }
 
-var defaultSuccessFilter = func(err error) bool {
-	return err == nil
-}
-
 type instrumentedCall struct {
 	scope   Scope
 	success Counter
@@ -50,17 +46,9 @@ type instrumentedCall struct {
 // Exec executes the given block of code, and records whether it succeeded or
 // failed, and the amount of time that it took
 func (c *instrumentedCall) Exec(f ExecFn) error {
-	return c.ExecWithFilter(f, defaultSuccessFilter)
-}
-
-// ExecWithFilter executes the given block of code, and records whether it succeeded or
-// failed based on the result of a custom filter (e.g. the filter could determine a bad request error
-// to be actually success for server logic), and the amount of time that it took
-func (c *instrumentedCall) ExecWithFilter(f ExecFn, isSuccess SuccessFilterFn) error {
 	sw := c.timing.Start()
 
-	err := f()
-	if err != nil && !isSuccess(err) {
+	if err := f(); err != nil {
 		c.error.Inc(1.0)
 		return err
 	}
@@ -68,5 +56,5 @@ func (c *instrumentedCall) ExecWithFilter(f ExecFn, isSuccess SuccessFilterFn) e
 	sw.Stop()
 	c.success.Inc(1.0)
 
-	return err
+	return nil
 }
--- a/instrument.go
+++ b/instrument.go
@@ -24,12 +24,19 @@ import (
 	"fmt"
 )
 
+const (
+	_resultType        = "result_type"
+	_resultTypeError   = "error"
+	_resultTypeSuccess = "success"
+	_timingFormat      = "%s.latency"
+)
+
 // NewInstrumentedCall returns an InstrumentedCall with the given name
 func NewInstrumentedCall(scope Scope, name string) InstrumentedCall {
 	return &instrumentedCall{
-		error:   scope.Tagged(map[string]string{"result_type": "error"}).Counter(name),
-		success: scope.Tagged(map[string]string{"result_type": "success"}).Counter(name),
-		timing:  scope.Timer(fmt.Sprintf("%s.latency", name)),
+		error:   scope.Tagged(map[string]string{_resultType: _resultTypeError}).Counter(name),
+		success: scope.Tagged(map[string]string{_resultType: _resultTypeSuccess}).Counter(name),
+		timing:  scope.Timer(fmt.Sprintf(_timingFormat, name)),
 	}
 }
 
--- a/instrument.go
+++ b/instrument.go
@@ -20,15 +20,11 @@
 
 package tally
 
-import (
-	"fmt"
-)
-
 const (
 	_resultType        = "result_type"
 	_resultTypeError   = "error"
 	_resultTypeSuccess = "success"
-	_timingFormat      = "%s.latency"
+	_timingFormat      = "latency"
 )
 
 // NewInstrumentedCall returns an InstrumentedCall with the given name
@@ -36,7 +32,7 @@ func NewInstrumentedCall(scope Scope, name string) InstrumentedCall {
 	return &instrumentedCall{
 		error:   scope.Tagged(map[string]string{_resultType: _resultTypeError}).Counter(name),
 		success: scope.Tagged(map[string]string{_resultType: _resultTypeSuccess}).Counter(name),
-		timing:  scope.Timer(fmt.Sprintf(_timingFormat, name)),
+		timing:  scope.SubScope(name).Timer(_timingFormat),
 	}
 }
 
--- a/instrument.go
+++ b/instrument.go
@@ -1,3 +1,23 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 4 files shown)


types.go
```diff
@@ -57,6 +57,9 @@ type Scope interface {
 
 	// Capabilities returns a description of metrics reporting capabilities.
 	Capabilities() Capabilities
+
+	// InstrumentedCall returns an InstrumentedCall with the given name
+	InstrumentedCall(name string) InstrumentedCall
```
robskillington: Adding this method to this interface is going to cause a major semver bump for tally which we are committed to not doing again.

The only way to provide this as a non-breaking change is to provide it as a module level constructor.  e.g. `func NewInstrumentedCall(scope Scope, name string) InstrumentedCall { ... }`.  

This is the route we'll need to go down to avoid a breaking(major semver) change.

robskillington: To make this API a little more readable we could create a new subpackage called `instrument` and export `func NewCall(scope tally.Scope, name string) Call { ... }`.  That way any future things we'd like to wrap for instrumenting purposes can live in this subpackage.  e.g. You could imagine a `func NewContextExecution(scope tally.Scope, name string) ContextExecution { ... }` that would implement something like:
```go
// public type
type ContextExecution interface {
  // RecordExec records the execution of a context that is cancelable, returns 
  // false if context is not cancelable and cannot be recorded, true if it can and will be
  RecordExec(ctx context.Context) bool
}

// implementation
type contextExecution struct {
  success, error tally.Counter
  latency tally.Timer
}

func NewContextExecution(scope tally.Scope, name string) ContextExecution {
  return &contextExecution{
    success: scope.Tagged(map[string]string{"result_type":"success"}).Counter(name),
    error: scope.Tagged(map[string]string{"result_type":"error"}).Counter(name),
    latency: scope.Timer(fmt.Sprintf("%s.latency", name)),
  }
}

func (c *contextExecution) RecordExec(ctx context.Context) bool {
  doneCh := ctx.Done()
  if doneCh == nil {
    return false // ctx not cancellable, can't record execution
  }
  start := time.Now()
  go func() {
    <-doneCh
    duration := time.Since(start)
    c.success.Inc(1)
    c.error.Inc(1)
    c.latency.Record(duration)
  }()
  return true
}
```

proflayton: I see. Good point. I wanted to keep in parity with go-common, but that isn't worth the major bump required.

types.go
```diff
@@ -103,6 +106,24 @@ type Stopwatch struct {
...
+	// failed, and the amount of time that it took
+	Exec(f func() error) error
+
+	// ExecWithFilter executes the given block of code, and records whether it succeeded or
+	// failed based on the result of a custom filter (e.g. the filter could determine a bad request error
+	// to be actually success for server logic), and the amount of time that it took
+	ExecWithFilter(f func() error, isSuccess SuccessFilter) error
+
+	// Tagged returns a new InstrumentedCall with the given set of tags
+	Tagged(tags map[string]string) InstrumentedCall
```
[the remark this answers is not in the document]
proflayton: Sounds good to me

types.go
```diff
@@ -150,3 +150,21 @@ type Capabilities interface {
 	// Tagging returns whether the reporter has the capability for tagged metrics.
 	Tagging() bool
 }
+
+// SuccessFilter determines whether an error should be considered success
+type SuccessFilter func(err error) bool
```
robskillington: nit: Can we standardize the  naming just a little and name this `SuccessFilterFn` to match the suffix of `ExecFn`?

instrument.go
```diff
@@ -0,0 +1,49 @@
...
+type instrumentedCall struct {
+	scope   Scope
+	success Counter
+	error   Counter
+	timing  Timer
+}
+
+// Exec executes the given block of code, and records whether it succeeded or
+// failed, and the amount of time that it took
+func (call *instrumentedCall) Exec(f ExecFn) error {
```
robskillington: nit: To standardize could you use a single letter for these receiver types, e.g. `func (c *instrumentedCall) ...`

scope_test.go
```diff
@@ -313,6 +314,43 @@ func TestWriteReportLoop(t *testing.T) {
...
 }
 
+func TestInstrumentedCallSuccess(t *testing.T) {
+	r := newTestStatsReporter()
+	s, closer := NewRootScope(ScopeOptions{Reporter: r}, 10)
+	defer closer.Close()
+
+	r.cg.Add(1)
+	r.tg.Add(1)
+	err := NewInstrumentedCall(s, "testInstrumented").Exec(func() error {
```
robskillington: nit: For consistency can you name these as underscore separated rather than lower camel case?  This matches the rest of the tests and examples in the project. i.e. `test_instrumented`, etc.

types.go
proflayton: Sounds good 👍

instrument.go
```diff
@@ -0,0 +1,69 @@
...
+// THE SOFTWARE.
+
+package tally
+
+import (
+	"fmt"
+)
+
+// NewInstrumentedCall returns an InstrumentedCall with the given name
+func NewInstrumentedCall(scope Scope, name string) InstrumentedCall {
```
prateek: Could you move this into a sub-package called `instrument`

instrument.go
```diff
@@ -0,0 +1,69 @@
...
+	return &instrumentedCall{
+		error:   scope.Tagged(map[string]string{"result_type": "error"}).Counter(name),
```
prateek: nit: use constants instead of inline strings

i.e. add the following to the top of the file 
```
const (
  resultTypeString = "result_type" 
) 
```

also, use a enum type for error/success, something like 
```
type resultType string 

const (
  resultTypeSuccess resultType = iota
  resultTypeError
)
```

instrument.go
proflayton: I was initially going to, but it makes testing really gross since it will rely on the testStatsReporter and that would cause circular dependencies.

instrument.go
robskillington: Won't be able to use an enum type as it tag values need to be a string, but you can use:
```go
type resultType string

const (
  successResultType resultType = "success"
  errorResultType resultType = "error"
)
```

And for the actual tag name probably best to name it as such (indicating its a tag name):
```go
const resultTypeTagName = "result_type"
```

proflayton: Sorry, I am failing to see the purpose of the enum type. Having `0` and `1` isn't very human readable for m3 tags?

instrument.go
robskillington: In regards to the commit description re: testing you can use `tally.NewTestScope()` and `Snapshot()` to capture the output of the scope instead of the private test reporter in the root tally package.

types.go
```diff
@@ -150,3 +150,21 @@ type Capabilities interface {
...
+// An InstrumentedCall allows tracking the success, errors, and timing of blocks of code
+type InstrumentedCall interface {
+	// Exec executes the given block of code, and records whether it succeeded or
+	// failed, and the amount of time that it took
+	Exec(f ExecFn) error
+
+	// ExecWithFilter executes the given block of code, and records whether it succeeded or
+	// failed based on the result of a custom filter (e.g. the filter could determine a bad request error
+	// to be actually success for server logic), and the amount of time that it took
+	ExecWithFilter(f ExecFn, isSuccess SuccessFilterFn) error
```
prateek: I'm not sure I follow the value of the additional method here. If needed, wouldn't it be as simple as a lambda wrapper around the `ExecFn`?

i.e. lets say you have some code using `ExecWithFilter` like so –
```
func execFn() error {
  // ... your exec fn
}

func successFilterFn(error) bool {
  // ... your filter fn
}


func XYZ() {
  instrumentedCall := ...
  instrumentedCall.ExecWithFilter(execFn, successFilterFn)
}
```

Isn't that the same as –

```
// other func defs 
func XYZ {
  instrumentedCall := ...
  instrumentedCall.Exec(func() error {
  err := execFn()
    if !successFilterFn(err) {
      return err
    }
  return nil
  })
}
```

instrument.go
prateek: sorry, meant to use the string version @robskillington suggested, `iota` was muscle memory

types.go
proflayton: I wanted to keep the interface the same as the go-common interface so when migrating you can use the same functions.

robskillington: You would have to use the new constructor either way however.  It does seem like not a fantastic API when the way @prateek pointed out is actually simpler and provides better readability in terms of users who don't know what the second lambda does exactly without reading the interface.

instrument.go
```diff
@@ -0,0 +1,76 @@
...
+	_resultTypeSuccess = "success"
+	_timingFormat      = "%s.latency"
+)
+
+// NewInstrumentedCall returns an InstrumentedCall with the given name
+func NewInstrumentedCall(scope Scope, name string) InstrumentedCall {
+	return &instrumentedCall{
+		error:   scope.Tagged(map[string]string{_resultType: _resultTypeError}).Counter(name),
+		success: scope.Tagged(map[string]string{_resultType: _resultTypeSuccess}).Counter(name),
+		timing:  scope.Timer(fmt.Sprintf(_timingFormat, name)),
```
robskillington: I should've mentioned this earlier, we should replace `scope.Timer(fmt.Sprintf(fmt, name))` with `scope.SubScope(name).Timer(timingSuffix)` so that it uses the user configured `Separator` string in the `ScopeOptions` provided to the scope.

proflayton: Oh sweet. Didn't know about that method. Sounds good!

PR #60, opened 2017-09-24
Unpin facebook/clock

kriskowal:
This change removes a hard pin on what is currently the master commit on facebook/clock. Like many Go projects, the project appears to have the release discipline of "never break master". With this pin, any project that depends on both tally and facebook/clock will need to use the exact same hash.

Alternately, I’d consider asking facebook/clock to cut a tagged semver release.

### DIFF
--- a/glide.yaml
+++ b/glide.yaml
@@ -6,7 +6,6 @@ import:
   subpackages:
   - statsd
 - package: github.com/facebookgo/clock
-  version: master
 - package: github.com/prometheus/client_golang
   version: ^0.8.0
   subpackages:
--- a/glide.yaml
+++ b/glide.yaml
@@ -6,7 +6,7 @@ import:
   subpackages:
   - statsd
 - package: github.com/facebookgo/clock
-  version: 600d898af40aa09a7a93ecb9265d87b0504b6f03
+  version: master
 - package: github.com/prometheus/client_golang
   version: ^0.8.0
   subpackages:


glide.yaml
```diff
@@ -6,7 +6,7 @@ import:
   subpackages:
   - statsd
 - package: github.com/facebookgo/clock
-  version: 600d898af40aa09a7a93ecb9265d87b0504b6f03
+  version: master
```
abhinav: Might be preferable to have no pin at all.

That way consumers of tally aren't forced to pin to master and can pin to a hash if that's what they need.

robskillington: I'd be in preference of just removing the dependency altogether.  

To mock a clock is really simple and we've done it tons of times just with a 20 line or so implementation.  You rarely need the extra features a package like this offers.

abhinav: Another option is to actually check in the vendored code for this library. facebookgo/clock doesn't cross Tally's API boundaries so github.com/uber-go/tally/vendor/github.com/facebookgo/clock being different from github.com/facebookgo/clock shouldn't be an issue.

This is something ThriftRW Go already does with github.com/fatih/structtag: https://github.com/thriftrw/thriftrw-go/tree/dev/vendor/github.com/fatih/structtag

kriskowal:
I’ve opted to just remove the pin entirely. It’s unlikely to break going forward. @robskillington I think it’s a great idea to remove the mock clock entirely. Let’s land this and track that work forward.

robskillington:
Hey Kris, sorry it's taken so long to get around to this.

I'll remove the dependency:
https://github.com/uber-go/tally/pull/76

PR #61, opened 2017-09-29
Change default histogram buckets from {-infinity,0} to {-infinity,first_defined}

katezaps:
When zero is present


histogram_test.go
```diff
@@ -86,7 +86,7 @@ func TestBucketPairsSortsDurationBuckets(t *testing.T) {
 	assert.Equal(t, time.Duration(math.MaxInt64), pairs[3].UpperBoundDuration())
 }
 
-func TestBucketPairsDurationBucketsInsertsMissingZero(t *testing.T) {
+func TestBucketPairsDurationBucketsDoesntInsertMissingZero(t *testing.T) {
```
robskillington: Probably can delete this test, it was mainly for branch coverage and verification of an added feature.  We don't really need to test the non-existence of features.

PR #65, opened 2017-10-23
Use go.uber.org/atomic ^1

abhinav:
Instead of relying on a pre-1.0 version of uber-go/atomic, this switches
to the 1.x series of go.uber.org/atomic.


abhinav:
@robskillington Is this good to merge?

prashantv:
Ping @robskillington -- currently tally pins to a specific commit hash in the `glide.yaml` which makes it harder for users to use it when other libraries want `^1`.

prashantv:
Ping again, want to remove pinned hashes from the `glide.yaml`.

robskillington:
I don't have a good mechanism for GitHub notifications unfortunately, apologies.

PR #66, opened 2017-11-14
[Stopwatch] Add IsValid method to check if a Stopwatch has been initialized properly

jeromefroe:
There are occasions where one may need to check if a `Stopwatch` has been initialized properly. One such example can be found in the [`m3x`](https://github.com/m3db/m3x/blob/master/instrument/methods.go#L66) package. However, since `Stopwatch` contains a `time.Time` field internally, comparing the given stopwatch to the zero value for a `Stopwatch` is insufficient. Consequently, this PR adds an `IsValid` method to `Stopwatch` to determine if it has been initialized with a recorder properly.

cc @robskillington @richardartoul

### DIFF
--- a/types.go
+++ b/types.go
@@ -114,6 +114,11 @@ func (sw Stopwatch) Stop() {
 	sw.recorder.RecordStopwatch(sw.start)
 }
 
+// IsValid reports whether the StopWatch is valid for use.
+func (sw Stopwatch) IsValid() bool {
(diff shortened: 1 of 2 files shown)


robskillington:
I think it'd be preferable to instead make stopping a stopwatch that is not initialized properly to be a no-op rather than make user defend in their code.

PR #70, opened 2017-12-07
Fix TestBatchSizes failing on Travis for M3 reporter

robskillington:


m3/reporter_test.go
```diff
@@ -383,6 +384,8 @@ func TestBatchSizes(t *testing.T) {
 			}
 		)
 		for atomic.LoadUint32(&stop) == 0 {
+			wg.Add(1)
```
prateek: don't entirely follow who's using `wg`. I can trace it to wg.Done in the fakeM3Service but nothing seems to be calling `wg.Wati()`.

robskillington: So basically this is only to avoid a negative wait counter in the fake m3 service that uses the wg.  I'll make this nil-able so we can remove this `wg.Add(1)` for the test that doesn't need to actually `wg.Wait()`.

PR #72, opened 2018-02-15
Reset metric tags after returning metric to pool

xichen2020:
cc @robskillington @jeromefroe @martin-mao

Currently if we allocate a new metric with no tags, either by calling `reporter.Allocate{Counter,Timer,Gauge,Histogram}` directly or calling `scope.{Counter,Timer,Gauge,Histogram}` on a scope with no tags, the new metric may reuse a pooled metric with its previous tags even though the new metric itself does not have tags, causing metrics to be associated with wrong tags.

This PR fixes the issue and adds a regression test that fails on master but passes with the PR.

### DIFF
--- a/example/main.go
+++ b/example/main.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2017 Uber Technologies, Inc.
+// Copyright (c) 2018 Uber Technologies, Inc.
 //
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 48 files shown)


xichen2020:
Lint's failing due to license header containing year 2017 instead of 2018, need to either save the 2017 license locally and run with `uber-licence --license license_file` or mass update all licenses. Either way unrelated to this PR.

robskillington:
Good find, I had a suspicion pooling was responsible somewhere... I'll fix the license and push a change.

xichen2020:
Hey thanks for merging, should we cut a new release so tally users can pick it up when they update dependencies?

xichen2020:
Brilliant thanks!

PR #73, opened 2018-02-22
add ExecWithFilter

klnusbaum:
Sometimes, an instrumented call can return an error that we don't actually want to count as an error, but rather as a success. A classic use case for this is returning an error like "BadRequest" to a caller. In this case, the server hasn't actually caused an error. It's running as expected and we'd like to count this as a success. Never the less, we still need to propagate the error to the caller.

Let's add `ExecWithFilter`, it's the same as `Exec`, except we allow the user to define a function that let's use know whether or not to log the result as an error or if the call was actually a success despite the error.

### DIFF
--- a/instrument/call.go
+++ b/instrument/call.go
@@ -66,11 +66,11 @@ func (c *call) Exec(f ExecFn) error {
 	return nil
 }
 
-func (c *call) ExecWithFilter(f ExecFn, success SuccessFilter) error {
+func (c *call) ExecWithFilter(f ExecFn, success FilterFn) error {
 	sw := c.timing.Start()
 	err := f()
 
-	if err := f(); err != nil && !success(err) {
+	if err := f(); !success(err) {
 		c.error.Inc(1.0)
 		return err
 	}
--- a/instrument/call.go
+++ b/instrument/call.go
@@ -65,3 +65,18 @@ func (c *call) Exec(f ExecFn) error {
 
 	return nil
 }
+
+func (c *call) ExecWithFilter(f ExecFn, success SuccessFilter) error {
+	sw := c.timing.Start()
+	err := f()
+
+	if err := f(); err != nil && !success(err) {
+		c.error.Inc(1.0)
+		return err
+	}
+
+	sw.Stop()
+	c.success.Inc(1.0)
+
+	return err
+}
--- a/instrument/call.go
+++ b/instrument/call.go
@@ -66,17 +66,19 @@ func (c *call) Exec(f ExecFn) error {
 	return nil
 }
 
-func (c *call) ExecWithFilter(f ExecFn, success FilterFn) error {
-	sw := c.timing.Start()
-	err := f()
-
-	if err := f(); !success(err) {
-		c.error.Inc(1.0)
-		return err
-	}
-
-	sw.Stop()
-	c.success.Inc(1.0)
-
+// ExecWithFilter executes a function and records whether it succeeded or
+// failed, and the amount of time that it took. ExecWithFilter
+// will always return the error associted with the called function.
+// However, it will decide whether or not classify that error as
+// a success or failure based on the provided SuccessFilter.
+func ExecWithFilter(call Call, f ExecFn, success FilterFn) error {
+	var err error
+	call.Exec(func() error {
+		err = f()
+		if !success(err) {
+			return err
+		}
+		return nil
+	})
 	return err
 }
--- a/instrument/types.go
+++ b/instrument/types.go
@@ -23,20 +23,21 @@ package instrument
 // ExecFn is an executable function that can be instrumented with a Call.
 type ExecFn func() error
 
-// SuccessFilter takes in an error and, if the error does not actually indicate
+// FilterFn takes in an error and, if the error does not actually indicate
 // a failure, returns true. If the error does actually indicate failure, the
 // function returns false.
-type SuccessFilter func(err error) bool
+type FilterFn func(err error) bool
 
 // Call allows tracking the successes, errors, and timing of functions.
 type Call interface {
 	// Exec executes a function and records whether it succeeded or
 	// failed, and the amount of time that it took.
 	Exec(f ExecFn) error
+
 	// Exec executes a function and records whether it succeeded or
 	// failed, and the amount of time that it took. ExecWithFilter
 	// will always return the error associted with the called function.
 	// However, it will decide whether or not classify that error as
 	// a success or failure based on the provided SuccessFilter.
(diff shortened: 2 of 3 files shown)


instrument/types.go
```diff
@@ -23,9 +23,20 @@ package instrument
 // ExecFn is an executable function that can be instrumented with a Call.
 type ExecFn func() error
 
+// SuccessFilter takes in an error and, if the error does not actually indicate
```
prateek: nit: could you rename to `FilterFn`

instrument/types.go
```diff
@@ -23,9 +23,20 @@ package instrument
...
+// a failure, returns true. If the error does actually indicate failure, the
+// function returns false.
+type SuccessFilter func(err error) bool
+
 // Call allows tracking the successes, errors, and timing of functions.
 type Call interface {
 	// Exec executes a function and records whether it succeeded or
 	// failed, and the amount of time that it took.
 	Exec(f ExecFn) error
+	// Exec executes a function and records whether it succeeded or
```
prateek: nit: insert new line here

instrument/call.go
```diff
@@ -65,3 +65,18 @@ func (c *call) Exec(f ExecFn) error {
 
 	return nil
 }
+
+func (c *call) ExecWithFilter(f ExecFn, success SuccessFilter) error {
+	sw := c.timing.Start()
+	err := f()
+
+	if err := f(); err != nil && !success(err) {
```
prateek: should this be:

```
if err := f(); !success(err) {
...
}
```

instrument/call.go
```diff
@@ -65,3 +65,18 @@ func (c *call) Exec(f ExecFn) error {
 
 	return nil
 }
+
```
prateek: could you re-implement `Exec` to use `ExecWithFilter` internally

klnusbaum: We absolutely could and I considered this. However, `Exec` is most likely the common case for most people, and if we implement it with `ExecWithFilter` we're now going to be incurring two extra function calls (one to `ExectWithFilter` itself, and then one to a default filter that always returns false). I'm not a performance expert, but I was worried about how this would affect performance for the common case.

Will Go just inline the function calls anyway? Want me to do any benchmarking? Or am I just over thinking things here and we should just implement `Exec` by calling `ExecWithFilter`?

prashantv:
Just a note -- adding a method to an interface is considered a breaking change in Go, so this would require a major version bump.

Given the pain that major version bumps cause, I'm not sure this is something that's worth adding. Is there another backwards compatible way to make this work?

instrument/call.go
klnusbaum: I wrote a quick benchmark and have some not too surprising results. I implemented an `Exec2` function that just calls `ExecWithFilter` as you suggested. I then wrote the following benchmarks (where `Exec` is just the current `Exec` left as is:

```go

func BenchmarkExec(b *testing.B) {
	s := tally.NewTestScope("", nil)
	instrumented := NewCall(s, "test_call")
	for n := 0; n < b.N; n++ {
		instrumented.Exec(func() error {
			if rand.Intn(1000) == 999 {
				return errors.New("expected")
			}
			return nil
		})
	}
}

func BenchmarkExec2(b *testing.B) {
	s := tally.NewTestScope("", nil)
	instrumented := NewCall(s, "test_call")
	for n := 0; n < b.N; n++ {
		instrumented.Exec2(func() error {
			if rand.Intn(1000) == 999 {
				return errors.New("expected")
			}
			return nil
		})
	}
}
```
Basically, call the instrumented function a bunch of times, and every so often have the function return an error (1 in a 1000 times to be exact).

Here are the results from three runs of the benchmark:
```
kurtis-C02T515QHF1T:instrument kurtisnusbaum$ go test -bench=.
goos: darwin
goarch: amd64
pkg: github.com/uber-go/tally/instrument
BenchmarkExec-4          1000000              1604 ns/op
BenchmarkExec2-4         1000000              1653 ns/op
PASS
ok      github.com/uber-go/tally/instrument     3.323s
kurtis-C02T515QHF1T:instrument kurtisnusbaum$ go test -bench=.
goos: darwin
goarch: amd64
pkg: github.com/uber-go/tally/instrument
BenchmarkExec-4          1000000              1587 ns/op
BenchmarkExec2-4         1000000              1613 ns/op
PASS
ok      github.com/uber-go/tally/instrument     3.265s
kurtis-C02T515QHF1T:instrument kurtisnusbaum$ go test -bench=.
goos: darwin
goarch: amd64
pkg: github.com/uber-go/tally/instrument
BenchmarkExec-4          1000000              1612 ns/op
BenchmarkExec2-4         1000000              1678 ns/op
PASS
ok      github.com/uber-go/tally/instrument     3.355s
```
As you can see, `Exec2` is just every so slightly slow in a pretty consistent fashion. It looks like on average `Exec` as is runs about 50 ns faster. I guess the question now is, do we value the code deduplication more than we value the 50 ns. Let me know what you think and I'll make the change.

klnusbaum:
Oh shoot, you're right Prashant. I suppose we could create a function like the following:
```go
func ExecWithFilter(call Call, f ExecFn, success Filter)
```
and just not add that to the interface. Kinda sad and hacky though :/

instrument/call.go
prateek: tl;dr - yeah I don't think the 50ns matter for this case. We're not anywhere nearly that frequency for metric sampling. Go ahead and rewrite into the new version.

Longer story –

I did a benchmark to see function execution overhead and it was pretty much nothing.
```
func call(err error) error {
    return err
}
 
func BenchmarkInlining(b *testing.B) {
    err := fmt.Errorf("random err")
    b.ReportAllocs()
    for i := 0; i < b.N; i++ {
        err = call(err)
    }
}
```

Results:
```
    goos: darwin
    goarch: amd64
    BenchmarkInlining-4     2000000000           0.33 ns/op        0 B/op          0 allocs/op
    PASS
 ```

There are a couple of things different about your benchmark than the final version would look like:
- allocate the passthru lambda on the stack, rather than inline 
- your lambda is doing things beyond just passthru

Would be interesting to see the results of your benchmark with these changes

klnusbaum:
I removed `ExecWithFilter` from the interface to prevent the need for a major version bump.

At this point, I'm not sure if the code is worth committing to Tally. It's code that anyone could write in their codebase should they need it. One day if we make a new major version of Tally, this might be worth it since we can include it as part of the `Call` interface. But as it is, this code seems not that compelling. Feel free to close the PR.

If you do find it compelling though, I'm happy to merge it.

mway:
Closing due to age. Feel free to reopen as necessary!

PR #75, opened 2018-05-09
Emit latency metric on both success and error for instrumented call

robskillington:
cc @kamran-m

### DIFF
--- a/instrument/call.go
+++ b/instrument/call.go
@@ -39,7 +39,7 @@ const (
 // {{name}}.latency
 func NewCall(scope tally.Scope, name string) Call {
 	return &call{
-		error:   scope.Tagged(map[string]string{resultType: resultTypeError}).Counter(name),
+		err:     scope.Tagged(map[string]string{resultType: resultTypeError}).Counter(name),
 		success: scope.Tagged(map[string]string{resultType: resultTypeSuccess}).Counter(name),
 		timing:  scope.SubScope(name).Timer(timingSuffix),
 	}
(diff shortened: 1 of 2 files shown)


prashantv:
> The subtlety of the instrumented call only emitting metrics on success can hide latency that occurs when an error is encountered. This change will emit latency in the case of errors as well to revert this subtle behavior.

There was a discussion on a similar change internally (D1537725), cc @xichen2020 who made some good points that this might be an unexpected behaviour change,
> existing users may see unexpected timing drops / spikes in their graphs once they pull in this change without knowing why.

instrument/call.go
```diff
@@ -48,20 +48,20 @@ func NewCall(scope tally.Scope, name string) Call {
...
 	scope   tally.Scope
 	success tally.Counter
-	error   tally.Counter
+	err     tally.Counter
 	timing  tally.Timer
 }
 
 func (c *call) Exec(f ExecFn) error {
 	sw := c.timing.Start()
+	err := f()
```
xichen2020: Would it be worth providing separate optional timing stats for success latencies and error latencies in addition to the overall timing stats? The overall timing stat may be insufficient if the user wants to gain additionally visibility into e.g., error latencies since that can be very different from the overall latency.

jeromefroe: If we separate the latencies based on success or error I think we would probably want to include a timer for both as well. Otherwise one wouldn't be able to calculate the true latency across all requests.

PR #80, opened 2018-06-15
Add support for specifying 'Gatherer' in prometheus Reporter options.

klueska:
Passing the Gatherer is required to be able return the HTTPHandler
appropriate for a customer Registerer that is passed in. Without this
change, the HTTPHandler is *always* associated with the Gatherer for the
promhttp.DefaultRegisterer, which is confusing if a custom Registerer is
passed.

### DIFF
--- a/prometheus/reporter.go
+++ b/prometheus/reporter.go
@@ -143,6 +143,7 @@ type metricID string
 type reporter struct {
 	sync.RWMutex
 	registerer      prom.Registerer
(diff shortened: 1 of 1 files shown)


robskillington:
Hey @klueska this makes sense, would you mind writing a test to verify that the gatherer option is respected perhaps?

robskillington:
I'm going to merge this, since it makes sense.

Issue #82, opened 2018-07-16
Tally should not pin any dependencies to hashes in glide.yaml


prashantv:
Currently, tally has many pins to hashes in the `glide.yaml` file:

E.g.,
```
- package: github.com/m3db/prometheus_client_golang
  version: 8ae269d24972b8695572fa6b2e3718b5ea82d6b4
```

```
- package: github.com/apache/thrift
  version: 9549b25c77587b29be4e0b5c258221a4ed85d37a
```

These are problematic for a few reasons:
 * It's not clear what API surface tally is pinning to, as it's not using a semver pin
 * It's very rigid, and likely to cause problems with another package that uses the same dependency, as they're unlikely to agree on the same hash
 * Dep ignores pins to specific revisions (Dep logs: `Ignoring pinned constraint 9549b25c77587b29be4e0b5c258221a4ed85d37a for github.com/apache/thrift." when importing uber-go/tally.`)

This causes problems for users as we migrate to dep, as the Tally constraints are ignored, and users end up getting the latest version of some dependencies (e.g., Thrift) which are not compatible (as they have a `ctx` parameter).

Can we ensure all non-test dependencies are pinned to version ranges where possible?

robskillington:
Honestly we don't care either way, we actually found pinning to hashes to be much less problematic with glide because it didn't try to strictly apply semver and was more relaxed (created less problems when trying to find a mutual version to select when conflicts were found between two dependencies using tally).

More than happy to move them to pinned ranges, I'd want to make sure our repositories can actually build with more strict ranges applied though before we actually merge to master.

robskillington:
We ended up pinning the remaining dependencies.

prashantv:
Perfect, thanks!

PR #83, opened 2018-07-17
Use semver dependency on Apache Thrift

bayesianmind:
Apache Thrift made a breaking api change in the 0.11 series. This package relies on the ~0.10 API. 
Currently downstream packages using dep do not inherit this package's transitive dependency constraint on apache thrift because dep does not consider revision pins when importing constraints from alternate dependency managers. By using a semver range instead of a direct pin.

Another parallel path is to propose amending Dep to consider Glide revision pins as valid transitive constraints. We can do that unilaterally in our fork of dep but it would be much better if we worked with upstream. This pull request gets us out of the immediate pain and unblocks migration to dep within Uber.

### DIFF
--- a/glide.yaml
+++ b/glide.yaml
@@ -16,7 +16,7 @@ import:
 - package: github.com/m3db/prometheus_procfs
   version: 1878d9fbb537119d24b21ca07effd591627cd160
 - package: github.com/apache/thrift
-  version: 9549b25c77587b29be4e0b5c258221a4ed85d37a
+  version: '>=0.9.3, <0.11.0'
   subpackages:
   - lib/go/thrift
 - package: go.uber.org/atomic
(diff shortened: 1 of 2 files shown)


prashantv:
Helps with #82, although there's many other dependencies in the `glide.yaml` that also should use semver ranges.

glide.yaml
```diff
@@ -16,7 +16,7 @@ import:
 - package: github.com/m3db/prometheus_procfs
   version: 1878d9fbb537119d24b21ca07effd591627cd160
 - package: github.com/apache/thrift
-  version: 9549b25c77587b29be4e0b5c258221a4ed85d37a
+  version: ~0.10
```
prashantv: Can this match tchannel-go's dependency?
`>=0.9.3, <0.11.0`
https://github.com/uber/tchannel-go/blob/dev/glide.yaml#L11

bayesianmind: Done.

robskillington:
LGTM to ease the pain for dep users.

Honestly we don't care either way, we actually found pinning to hashes to be much less problematic with `glide` because it didn't try to strictly apply semver and was more relaxed (created less problems when trying to find a mutual version to select when conflicts were found between two dependencies using tally).

PR #85, opened 2018-08-13
Add ReportableScope interface for use with pull interface

Intar:
You can use ReportableScope if you need to send data after some event happened.
For example, you can use this interface in HTTP Handler for pull-based metrics model.


robskillington:
@mway thoughts on this?  Expanding the API scope is probably not ideal, wonder if there's a similar way to support this without creating another interface.

mway:
> @mway thoughts on this? Expanding the API scope is probably not ideal, wonder if there's a similar way to support this without creating another interface.

agreed - we want scope convergence, not divergence. most of these types of problems can be solved outside of the API layer.

for example, we could hooks to the config struct and internalize it, but we'd need to bench the impact of adding hook evaluations (even noops) to the hot path. we shouldn't create use-case-specific hooks/callbacks, though - this should be solved generally.

will be gathering all of these types of use cases (this isn't altogether dissimilar from context baggage extraction that we've been talking about) as part of the conversations around making changes to the API.

it'd also be helpful to understand the impact so that we can prioritize accordingly.

PR #86, opened 2018-08-15
[prometheus-reporter] Add ability to infer listen port based on env var

prateek:
Ported from https://github.com/m3db/m3x/pull/186

### DIFF
--- a/prometheus/config.go
+++ b/prometheus/config.go
@@ -38,6 +38,11 @@ type Configuration struct {
 	// handler on the default HTTP serve mux without listening.
 	ListenAddress string `yaml:"listenAddress"`
 
+	// DynamicListenAddress if specified will be used instead of just registering the
+	// handler on the default HTTP serve mux without listening.
+	// Note: if DynamicListenAddress is specified, ListenAddress is ignored.
+	DynamicListenAddress *ListenAddressConfiguration `yaml:"dynamicListenAddress"`
+
 	// TimerType is the default Prometheus type to use for Tally timers.
 	TimerType string `yaml:"timerType"`
 
@@ -129,7 +134,12 @@ func (c Configuration) NewReporter(
 		path = handlerPath
 	}
 
-	if addr := strings.TrimSpace(c.ListenAddress); addr == "" {
+	addr, resolved, err := c.resolveListenAddress()
+	if err != nil {
+		return nil, err
+	}
+
+	if !resolved {
 		http.Handle(path, reporter.HTTPHandler())
 	} else {
 		mux := http.NewServeMux()
@@ -142,4 +152,23 @@ func (c Configuration) NewReporter(
 	}
 
 	return reporter, nil
+}
+
+func (c Configuration) resolveListenAddress() (addr string, resolved bool, err error) {
+	// first try DynamicListenAddress
+	if c.DynamicListenAddress != nil {
+		addr, err := c.DynamicListenAddress.Resolve()
+		if err != nil {
+			return "", false, err
+		}
+		return addr, true, nil
+	}
+
+	// next, try ListenAddress
+	addr = strings.TrimSpace(c.ListenAddress)
+	if addr == "" {
+		return "", false, nil
+	}
+
+	return addr, true, nil
 }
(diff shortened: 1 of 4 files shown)


prometheus/listen_config.go
```diff
@@ -0,0 +1,98 @@
...
+	"os"
+	"strconv"
+)
+
+// ListenAddressResolver is a type of port resolver
+type ListenAddressResolver string
+
+const (
+	// ConfigResolver resolves port using a value provided in config
+	ConfigResolver ListenAddressResolver = "config"
```
xichen2020: Would it be more clear to name this `PortType` with `static` and `dynamic` being the two values? Static ports should always be loaded from config, and dynamic ports should always be resolved from env vars.

prometheus/config.go
```diff
@@ -38,6 +38,11 @@ type Configuration struct {
 	// handler on the default HTTP serve mux without listening.
 	ListenAddress string `yaml:"listenAddress"`
 
+	// DynamicListenAddress if specified will be used instead of just registering the
+	// handler on the default HTTP serve mux without listening.
+	// Note: if DynamicListenAddress is specified, ListenAddress is ignored.
+	DynamicListenAddress *ListenAddressConfiguration `yaml:"dynamicListenAddress"`
```
xichen2020: Hm this is kinda confusing, e.g., you can have a `dynamicListenAddress` that defines its port in the config file.

prometheus/config.go
```diff
@@ -143,3 +153,22 @@ func (c Configuration) NewReporter(
 
 	return reporter, nil
 }
+
+func (c Configuration) resolveListenAddress() (addr string, resolved bool, err error) {
```
xichen2020: Technically you don't really need `resolved` since `addr == "" && !resolved` is always false.

prometheus/listen_config.go
```diff
@@ -0,0 +1,98 @@
...
+	Value *int `yaml:"value"`
+
+	// EnvVarName is the environment specified port if using environment port type.
+	EnvVarName *string `yaml:"envVarName"`
+}
+
+// Resolve returns the resolved listen address given the configuration.
+func (c ListenAddressConfiguration) Resolve() (string, error) {
+	if c.Port == nil {
+		return "", errors.New("missing port config")
```
xichen2020: nit: make the err `errMissingPort`.

prometheus/listen_config.go
```diff
@@ -0,0 +1,98 @@
...
+	EnvironmentResolver ListenAddressResolver = "environment"
+)
+
+// ListenAddressConfiguration is the configuration for resolving a listen address.
+type ListenAddressConfiguration struct {
+	// Hostname is the hostname to use
+	Hostname string `yaml:"hostname" validate:"nonzero"`
+
+	// port specifies the port to listen on
+	Port *PortConfiguration `yaml:"port" validate:"nonzero"`
```
xichen2020: Hm if this is always nonzero, why not make it a non-pointer?

prometheus/listen_config.go
```diff
@@ -0,0 +1,98 @@
...
+			err := fmt.Errorf("invalid port env var value using: resolver=%s, name=%s",
+				string(p.PortType), *p.EnvVarName)
+			return "", err
+		}
+	default:
+		return "", fmt.Errorf("unknown port type: resolver=%s",
+			string(p.PortType))
+	}
+
+	return fmt.Sprintf("%s:%d", c.Hostname, port), nil
```
xichen2020: nit: `net.JoinHostPort` might be more portable.

jeromefroe:
I thought about this some more, instead of adding another configuration option, how would you feel if we modify the prometheus configuration in the service itself? For example, we could change:

```
server.Run(server.RunOptions{
		ConfigFile: *configFile,
})
```

to

```
var cfg config.Configuration
if err := xconfig.LoadFile(&cfg, *configFile, xconfig.Options{}); err != nil {
			fmt.Fprintf(os.Stderr, "unable to load %s: %v", runOpts.ConfigFile, err)
			os.Exit(1)
}

// update config.Configuration.Metrics.PrometheusReporter based on env variable

server.Run(server.RunOptions{
		Config: config,
})
```

I think this approach would address both of our concerns: we wouldn't need to munge the config files at runtime and we wouldn't need to include an additional configuration option in tally.

Issue #89, opened 2019-02-20
Race condition when closing m3 reporter while metric reporting isn't finished


ChuntaoLu:
The code path [here](https://github.com/uber-go/tally/blob/master/m3/reporter.go#L432-L443)  causes potential race condition when metric reporting and closing reporter are in separate goroutines, i.e., one goroutine attempts to send to channel while another goroutine attempts to close the channel. 

With this behavior, applications that use tally experience flakiness when running tests (which involve closing reporter, e.g., server shutdown) with the race detector on.

PR #90, opened 2019-02-20
Use RWMutex instead of defer/recover to avoid publishing to closed channel in M3 reporter

ChuntaoLu:
This PR uses a close channel and a close signal metric to coordinate the closing of m3 reporter instead of the current panic-recovering approach, fixes https://github.com/uber-go/tally/issues/89.

There is potential race condition when reporting metric and closing reporter happen concurrently in separate goroutines, which is illustrated by this [test](https://github.com/uber-go/tally/compare/master...ChuntaoLu:lu.race?expand=1#diff-3e21eed7473a2862cd0db36ccbc836c4R198). The metric channel is a MPSC channel and the closer returned to user is likely in a goroutine that is neither a producer nor a consumer of the channel, therefore is not in a position to close the channel. This PR uses a close channel (SPMC where the closer goroutine is the sender and has the right to close the channel safely) to let the metric producers stop sending metrics and a signal metric to let the consumer stop processing metrics, so that we no longer need to close the channel, therefore avoid the race condition without explicit locks. 

Regarding the performance considerations in https://github.com/uber-go/tally/pull/46, below are the benchmark results:
```
➜  tally git:(lu.race) go test -bench BenchmarkTimer github.com/uber-go/tally/m3                                                     
goos: darwin
goarch: amd64
pkg: github.com/uber-go/tally/m3
BenchmarkTimer-8   	 2000000	       670 ns/op
PASS
ok  	github.com/uber-go/tally/m3	2.136s

➜  tally git:(master) go test -bench BenchmarkTimer github.com/uber-go/tally/m3                                                      
goos: darwin
goarch: amd64
pkg: github.com/uber-go/tally/m3
BenchmarkTimer-8   	 3000000	       680 ns/op
PASS
ok  	github.com/uber-go/tally/m3	2.739s
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,5 +30,4 @@ _testmain.go
 
 .DS_Store
 node_modules/
-
-
+.idea/
--- a/Makefile
+++ b/Makefile
@@ -27,7 +27,7 @@ lint:
 	@echo "Installing test dependencies for vet..."
 	@go test -i $(PKGS)
 	@echo "Checking vet..."
-	@$(foreach dir,$(PKG_FILES),go tool vet $(dir) 2>&1 | grep -v $(LINT_IGNORE) | tee -a lint.log;)
+	@$(foreach dir,$(PKG_FILES),go vet $(dir) 2>&1 | grep -v $(LINT_IGNORE) | tee -a lint.log;)
 	@echo "Checking lint..."
 	@$(foreach dir,$(PKGS),golint $(dir) 2>&1 | grep -v $(LINT_IGNORE) | tee -a lint.log;)
 	@echo "Checking for unresolved FIXMEs..."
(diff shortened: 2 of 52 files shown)


robskillington:
Thanks for the discovery and the fix.

I updated this to just a standard RW lock for protecting against sending on closed channel since the benchmark showed similar performance and the implementation is more idiomatic.

```
m3[lu.race]$ go test -v -bench BenchmarkTimer -benchtime 4s -run none .
goos: darwin
goarch: amd64
pkg: github.com/uber-go/tally/m3
BenchmarkTimer-4        10000000               726 ns/op
PASS
ok      github.com/uber-go/tally/m3     8.026s

m3[master]$ go test -v -bench BenchmarkTimer -benchtime 4s -run none .
goos: darwin
goarch: amd64
pkg: github.com/uber-go/tally/m3
BenchmarkTimer-4        10000000               699 ns/op
PASS
ok      github.com/uber-go/tally/m3     7.689s
```

PR #94, opened 2019-03-15
enable the tenancy for tally

lingfengZhang1989:
This issue is going to enable the tenancy for tally go library.

We extract the tenancy value from the jaeger baggage which is inside the context. Then we check the tenancy value (either "production" or "testing"), and set a tenancy tag in the tally scope.

We modified the Tagged() function to pass an optional context as parameter, and also provided a TaggedWithContext() function with context as one restrict parameter.

### DIFF
--- a/scope.go
+++ b/scope.go
@@ -21,12 +21,7 @@
 package tally
 
 import (
-	"encoding/base64"
-	"encoding/json"
-	"errors"
 	"fmt"
-	"context"
-	"github.com/opentracing/opentracing-go"
 	"io"
 	"sync"
 	"time"
@@ -62,11 +57,7 @@ var (
 const (
 	TestingTenancy     = "testing"
 	ProductionTenancy  = "production"
-	EmptyTenancy       = ""
-	ProductionTenancyString = "uber/production"
-	TestingTenancyString    = "uber/testing"
-	EmptyTenancyString      = ""
-	TenancyKey              = "request-tenancy"
+	TenancyKey         = "request-tenancy"
 )
 type scope struct {
 	separator      string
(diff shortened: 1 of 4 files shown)


scope.go
```diff
@@ -375,7 +383,27 @@ func (s *scope) Histogram(name string, b Buckets) Histogram {
 	return val
 }
 
-func (s *scope) Tagged(tags map[string]string) Scope {
+func (s *scope) Tagged(tags map[string]string, tenancys ...string) Scope {
```
mway: We should probably put `Tenancy string` in `ScopeOptions` such that it must be specified during `Scope` instantiation (if at all), as tenancy is something that would be fundamental to the scope of all metrics emitted by a given instance of `Scope`. In that model, we can just inject the tenancy into the tags upon instantiation and have that carried on with all subsequent `Scope` operations.

If you have a use case that requires a `Scope` that spans multiple tenancies (whether a metric that applies to N tenancies, or having a single root `Scope` have disparately-tenancied leaf `Scope`s), we should set up a chat so that I can learn more about your specific goals. I'm not opposed to adding support for both modes (both root `Scope` and leaf `Scope`), but I suspect the former will be far more prevalent and I'd like to start there.

types.go
```diff
@@ -50,7 +50,10 @@ type Scope interface {
 	Histogram(name string, buckets Buckets) Histogram
 
 	// Tagged returns a new child scope with the given tags and current tags.
-	Tagged(tags map[string]string) Scope
+	Tagged(tags map[string]string, tenancys ...string) Scope
+
+	// TaggedWithTenancy required the context to extract the tenancy value, return a new child scope with tenancy tag plus given tags
+	TaggedWithTenancy(tenancy string, tags map[string]string) Scope
```
prashantv: this technically is a backwards incompatible change:
https://blog.merovius.de/2015/07/29/backwards-compatibility-in-go.html
see "Adding a method to an interface"

which will require a new major release.

The functionality could be implemented without changing the interface (e.g., `tally.WithTenancy(scope tally.Scope, tenancy... string) tally.Scope` which would avoid a breaking change.

Another simpler option is just to have users add the tag themselves, and export a tenancy tag name.

mway: Fair point, although I'm not yet sure if we're going to observe the struct-fields-could-break-embedded-structs concern. Regardless - I think we should keep tenancy rooted, so that it's not something that suddenly appears somewhere in the chain of scopes at runtime - in other words, I like the idea of creating a subscope with tenancy less than I like creating the root scope with a tenancy in the first place. If you're going to use tenancy with your scope, it should be included all the way down. Thus, maybe `NewRootScopeWithTenancy(ScopeOptions, time.Duration, string)`?

scope.go
lingfengZhang1989: Hi Matt, I will set up a zoom chat regarding to the design requirements. Thanks!

PR #95, opened 2019-03-30
Add expiration functionality to metrics and scopes

martin-mao:
Currently, all created scopes and metrics (counters, gauges, timers and histograms) will live forever and all metrics will be checked in every report interval. For long lived  process use cases where short lived metrics are created commonly, the memory usage will grow over time with no mechanism to delete metrics that are no longer in use since we hold a reference to them in the scopes and registries.

This change adds a configurable expiration time in which scopes and metrics will be removed from the registry. The change is complicated by the fact that users may have their own references to these scopes and metric objects, which means that even if we removed our reference from the registry and scopes, we cannot safely clean these objects up (place them back into pools etc). Furthermore, if a user updates one of these metrics after it has expired, we need to re-insert that metric (and parent scope) back into the registry in order for the metric to be flushed successfully.

The overall logic is as follows:

- Scopes will only expire after the expiration period has elapsed and all of it's metrics have expired. Upon expiry, a scope will be removed from the registry.
- If a user holds onto a reference of an expired scope and calls to create a subscope or metric, we will check to see if another scope with the same prefix and tags has already been created and inserted into the registry in the meantime. If so, we will add subscopes and metrics to the new scope instead. If not, we will re-insert the scope back into the registry. There will only ever be one scope per name and tag combination in the registry at once, even though many may exist due to expiration and held references. This way, expired scopes will never contain non expired metrics and subscopes.
- Metrics will be expired if there have been no updates or increment calls for the duration of the expiration period. This check is applied during a report call to avoid a performance hit on the increment or update calls. Expired metrics will be removed from their parent scopes and will be GC'd as long as the user does not old a direct reference to them.
- If a user holds a direct reference to an expired metric and calls an increment or update, we will reinsert the metric to the appropriate scope in the registry and start reporting its value once again.
- If a scope has no reporter, the metrics will not be expired. 

There are no interface changes and all existing tests continue to pass.

### DIFF
--- a/scope.go
+++ b/scope.go
@@ -313,11 +313,11 @@ func (s *scope) cachedReport(c CachedStatsReporter) bool {
 }
 
 func (s *scope) clearExpiredMetrics(tracker *scopeExpiryTracker) {
-	var i int
-
 	if len(tracker.expiredCounters) > 0 {
 		s.cm.Lock()
 		for name, counters := range s.counters {
+			var i int
+
 			for _, counter := range counters {
 				if _, exists := tracker.expiredCounters[counter]; !exists ||
 					(exists && !counter.expired()) {
@@ -331,8 +331,6 @@ func (s *scope) clearExpiredMetrics(tracker *scopeExpiryTracker) {
 			} else {
 				s.counters[name] = counters[:i]
 			}
-
-			i = 0
 		}
 
 		tracker.numCounters = len(s.counters)
@@ -342,6 +340,8 @@ func (s *scope) clearExpiredMetrics(tracker *scopeExpiryTracker) {
 	if len(tracker.expiredGauges) > 0 {
 		s.gm.Lock()
 		for name, gauges := range s.gauges {
+			var i int
+
 			for _, gauge := range gauges {
 				if _, exists := tracker.expiredGauges[gauge]; !exists ||
 					(exists && !gauge.expired()) {
@@ -355,8 +355,6 @@ func (s *scope) clearExpiredMetrics(tracker *scopeExpiryTracker) {
 			} else {
 				s.gauges[name] = gauges[:i]
 			}
-
-			i = 0
 		}
 
 		tracker.numGauges = len(s.gauges)
@@ -366,6 +364,8 @@ func (s *scope) clearExpiredMetrics(tracker *scopeExpiryTracker) {
 	if len(tracker.expiredHistograms) > 0 {
 		s.hm.Lock()
 		for name, histograms := range s.histograms {
+			var i int
+
 			for _, histogram := range histograms {
 				if _, exists := tracker.expiredHistograms[histogram]; !exists ||
 					(exists && !histogram.expired()) {
@@ -379,8 +379,6 @@ func (s *scope) clearExpiredMetrics(tracker *scopeExpiryTracker) {
 			} else {
 				s.histograms[name] = histograms[:i]
 			}
-
-			i = 0
 		}
 
 		tracker.numHistograms = len(s.histograms)
@@ -458,6 +456,8 @@ func (s *scope) reportRegistryWithLock() {
 
 				if mets == 0 {
 					delete(s.registry.subscopes, name)
+				} else {
+					atomic.StoreUint32(&ss.expired, 0)
 				}
 			}
 		}
@@ -792,7 +792,7 @@ func (s *scope) Snapshot() Snapshot {
 							values[k] = 0
 						}
 
-						values[k] = values[k] + v
+						values[k] += v
 					}
 
 					for k, v := range histograms[i].snapshotDurations() {
@@ -800,7 +800,7 @@ func (s *scope) Snapshot() Snapshot {
 							durations[k] = 0
 						}
 
-						durations[k] = durations[k] + v
+						durations[k] += v
 					}
 				}
 
--- a/scope.go
+++ b/scope.go
@@ -204,11 +204,12 @@ func newRootScope(opts ScopeOptions, interval time.Duration) *scope {
 // report dumps all aggregated stats into the reporter. Should be called automatically by the root scope periodically.
 func (s *scope) report(r StatsReporter) bool {
 	tracker := newScopeExpiryTracker()
+	nowUnix := globalNow().Unix()
 
 	s.cm.RLock()
 	for name, counters := range s.counters {
 		for _, counter := range counters {
-			if counter.report(s.fullyQualifiedName(name), s.tags, r) {
+			if counter.report(s.fullyQualifiedName(name), s.tags, r, nowUnix) {
 				tracker.expiredCounters[counter] = struct{}{}
 			}
 		}
@@ -220,7 +221,7 @@ func (s *scope) report(r StatsReporter) bool {
 	s.gm.RLock()
 	for name, gauges := range s.gauges {
 		for _, gauge := range gauges {
-			if gauge.report(s.fullyQualifiedName(name), s.tags, r) {
+			if gauge.report(s.fullyQualifiedName(name), s.tags, r, nowUnix) {
 				tracker.expiredGauges[gauge] = struct{}{}
 			}
 		}
@@ -236,7 +237,7 @@ func (s *scope) report(r StatsReporter) bool {
 	s.hm.RLock()
 	for name, histograms := range s.histograms {
 		for _, histogram := range histograms {
-			if histogram.report(s.fullyQualifiedName(name), s.tags, r) {
+			if histogram.report(s.fullyQualifiedName(name), s.tags, r, nowUnix) {
 				tracker.expiredHistograms[histogram] = struct{}{}
 			}
 		}
@@ -258,11 +259,12 @@ func (s *scope) report(r StatsReporter) bool {
 
 func (s *scope) cachedReport(c CachedStatsReporter) bool {
 	tracker := newScopeExpiryTracker()
+	nowUnix := globalNow().Unix()
 
 	s.cm.RLock()
 	for _, counters := range s.counters {
 		for _, counter := range counters {
-			if counter.cachedReport() {
+			if counter.cachedReport(nowUnix) {
 				tracker.expiredCounters[counter] = struct{}{}
 			}
 		}
@@ -274,7 +276,7 @@ func (s *scope) cachedReport(c CachedStatsReporter) bool {
 	s.gm.RLock()
 	for _, gauges := range s.gauges {
 		for _, gauge := range gauges {
-			if gauge.cachedReport() {
+			if gauge.cachedReport(nowUnix) {
 				tracker.expiredGauges[gauge] = struct{}{}
 			}
 		}
@@ -290,7 +292,7 @@ func (s *scope) cachedReport(c CachedStatsReporter) bool {
 	s.hm.RLock()
 	for _, histograms := range s.histograms {
 		for _, histogram := range histograms {
-			if histogram.cachedReport() {
+			if histogram.cachedReport(nowUnix) {
 				tracker.expiredHistograms[histogram] = struct{}{}
 			}
 		}
--- a/scope.go
+++ b/scope.go
@@ -203,7 +203,7 @@ func newRootScope(opts ScopeOptions, interval time.Duration) *scope {
 
 // report dumps all aggregated stats into the reporter. Should be called automatically by the root scope periodically.
 func (s *scope) report(r StatsReporter) bool {
-	tracker := newMetricExpiryTracker()
+	tracker := newScopeExpiryTracker()
 
 	s.cm.RLock()
 	for name, counters := range s.counters {
@@ -257,7 +257,7 @@ func (s *scope) report(r StatsReporter) bool {
 }
 
 func (s *scope) cachedReport(c CachedStatsReporter) bool {
-	tracker := newMetricExpiryTracker()
+	tracker := newScopeExpiryTracker()
 
 	s.cm.RLock()
 	for _, counters := range s.counters {
@@ -310,7 +310,7 @@ func (s *scope) cachedReport(c CachedStatsReporter) bool {
 	return false
 }
 
-func (s *scope) clearExpiredMetrics(tracker *metricExpiryTracker) {
+func (s *scope) clearExpiredMetrics(tracker *scopeExpiryTracker) {
 	var i int
 
 	if len(tracker.expiredCounters) > 0 {
@@ -318,7 +318,7 @@ func (s *scope) clearExpiredMetrics(tracker *metricExpiryTracker) {
 		for name, counters := range s.counters {
 			for _, counter := range counters {
 				if _, exists := tracker.expiredCounters[counter]; !exists ||
-					(exists && atomic.LoadUint32(&counter.expired) == 0) {
+					(exists && !counter.expired()) {
 					counters[i] = counter
 					i++
 				}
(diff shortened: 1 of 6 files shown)


scope.go
```diff
@@ -182,52 +202,164 @@ func newRootScope(opts ScopeOptions, interval time.Duration) *scope {
...
 
 // report dumps all aggregated stats into the reporter. Should be called automatically by the root scope periodically.
-func (s *scope) report(r StatsReporter) {
-	s.cm.RLock()
-	for name, counter := range s.counters {
-		counter.report(s.fullyQualifiedName(name), s.tags, r)
+func (s *scope) report(r StatsReporter) bool {
+	var numMetrics, i int
+
+	s.cm.Lock()
```
prashantv: is `Lock` vs `RLock` going to have any negative impact?

should we instead under the rlock, come up with a list of counters to delete, and then do a separate `Lock` to just delete those items?

scope.go
```diff
@@ -55,27 +59,33 @@ var (
...
+	reporter            StatsReporter
+	cachedReporter      CachedStatsReporter
+	baseReporter        BaseStatsReporter
+	defaultBuckets      Buckets
+	sanitizer           Sanitizer
+	expirePeriodSeconds int64
+
+	registry    *scopeRegistry
+	status      scopeStatus
+	createdUnix int64
```
prashantv: is there an efficiency win from storing unix seconds rather than storing `*time.Time` and comparing using `.After`?

scope.go
```diff
@@ -261,118 +393,206 @@ func (s *scope) reportLoopRun() {
...
+// hasExpired indicates whether the scope is expired
+func (s *scope) hasExpired() bool {
+	return atomic.LoadUint32(&s.expired) == 1
+}
+
+// getCurrentScope gets the current active scope which could
+// be this scope or another if another was created after this
+// one expired
+func (s *scope) getCurrentScope() *scope {
+	if s.hasExpired() {
```
prashantv: could the counters/gauges/etc have the same logic, check if they've expired, and if so, get a new counter, and store a pointer to that counter? that could avoid storing `[]*counter` 

not sure if there's a huge benefit either way, but might simplify the code if a single name doesn't have multiple counters

scope.go
```diff
@@ -261,118 +393,206 @@ func (s *scope) reportLoopRun() {
...
+		// scope has expired. Go grab the current one from the registry
+		// TODO(martinm) - cache the registryKey in scope if expired occurs often
+		registryName := scopeRegistryKey(s.prefix, s.tags)
+		s.registry.RLock()
+		curScope, exists := s.registry.subscopes[registryName]
+		s.registry.RUnlock()
+
+		if exists {
```
prashantv: is it possible the scope we just got out has also expired?

stats.go
```diff
@@ -57,81 +57,244 @@ func (c *capabilities) Tagging() bool {
...
+		// was incremented, therefore insert back into scope
+		scope := c.scope.getCurrentScope()
+
+		scope.cm.Lock()
+		if counters, ok := scope.counters[c.name]; !ok {
+			// No counters with the same name were created since expiry
+			scope.counters[c.name] = []*counter{c}
+		} else {
+			exists := false
+			for _, counter := range counters {
```
prashantv: can avoid this check if instead, you did (after the if expired check):

```go
if !atomic.CompareAndSwapUint32(&c.expired, 1, 0) {
  // lost the race, another goroutine already added this counter back to the scope
  return
}

// now assume that we're not racing against ourself to add counter to `scope.counters[c.name]`

File: stats.go [hunk 6]
```diff
@@ -57,81 +57,244 @@ func (c *capabilities) Tagging() bool {
...
 	}
 
-	r.ReportCounter(name, tags, delta)
+	// Check if counter has expired
+	if c.scope != nil && globalNow().Unix() >
+		atomic.LoadInt64(&c.lastUpdateUnix)+c.scope.registry.expirePeriodSeconds {
+		atomic.StoreUint32(&c.expired, 1)
+		// Check once more for possible race where value updated between
+		// beginning of this call
+		delta = c.value()
```
prashantv: maybe move the logic of report + return if non-zero into an anonymous function to avoid duplicating the logic

stats.go
```diff
@@ -57,81 +57,244 @@ func (c *capabilities) Tagging() bool {
...
-	if delta == 0 {
-		return
+	if delta != 0 {
+		c.cachedCount.ReportCount(delta)
+		atomic.StoreInt64(&c.lastUpdateUnix, globalNow().Unix())
+		return false
+	}
+
+	// Check if counter has expired
+	if c.scope != nil && globalNow().Unix() >
```
prashantv: all of the logic to check whether we've expired, updating the expiry etc is common to `report` `cachedReport` for all the metric types.

i think you can probably create a embedded type that can track all the expiry metadata, and embed that in all the types so you're not duplicating the logic

scope.go
```diff
@@ -35,6 +36,9 @@ var (
 
 	globalNow = time.Now
 
+	// defaultExpiry is the default expiry period for scopes and metrics
+	defaultExpiry = 11 * time.Minute
```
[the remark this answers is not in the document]
martin-mao: Yep, figured it should be slightly above something periodic.

scope.go
martin-mao: Yeah about 20% difference in performance. Figured the logic wasn't that hard to read so worth the perf win.

scope.go
martin-mao: Yeah I thought about doing that - the main negative impact is going to be during a report call and users doing things like scope.Counter("myCounter").Inc(1) since those will be blocked, but those calls are already taking a complicated key generation and map lookup cost and are in the microseconds already. For users who care about perf, they're already holding a ref to the counter itself, so won't be impacted by this lock.

The problem with doing RLock first and deleting counters later is that it introduces a race condition such that you can have a metric marked as expired from the report call in line 211 and inserted into the to be deleted slice, and then marked as non expired if it gets updated before the Lock is acquired. This could result in a situation where the increment is lost.

I guess we can do a final check for expiration of the counter before doing the delete in the lock and if it's no longer expired, skip it. Will change.

scope.go
martin-mao: Yeah that'll be ideal, but unfortunately it can't be avoided. One can hold a ref to a counter and have it expire (and removed from the current/active scope) and then someone can create a new counter with the same name/tags combination. If they increment both, we now have two references to two counters which both need to be reported, so they have to be stored in a slice of counters.

scope.go
martin-mao: The logic to expire scopes happens within a s.registry.Lock() so when we look it up in line 434, it should exist. It could of course expire right after the RUnlock(), but that's why we do a final check before we return a metric from a counter/gauge/timer call to ensure the scope hasn't expired. If it has, we call the same function again to get a fresh scope.

stats.go
martin-mao: Yeah I can try and clean this up a bit, but having the two reports and them working completely differently has always been messy =P

stats.go
martin-mao: Yeah good call. I can also use:
if atomic.CompareAndSwapUint32(&c.expired, 1, 0) instead of if atomic.LoadUint32(&c.expired) == 1 in the original check, but CompareAndSwapUint32 is less performant than LoadUint32 :(

stats.go
```diff
@@ -355,6 +460,26 @@ func (h *histogram) RecordValue(value float64) {
 	// we always have a math.MaxFloat64 bucket.
 	idx := sort.SearchFloat64s(h.lookupByValue, value)
 	h.buckets[idx].samples.Inc(1)
+	h.checkExpiryAndUpdate()
+}
+
+func (h *histogram) checkExpiryAndUpdate() {
+	if h.tracker.expiredAndUpdated() {
```
prashantv: nit: return early 
```go
if !h.tracker.expiredAndUpdated() {
  return
}

// existing logic
```

stats.go
```diff
@@ -438,6 +568,44 @@ func newHistogramBucket(
...
+	name           string
+}
+
+// expiredAndUpdated checks the metric is expired and if so
+// will return true only for the first routine that updates
+// an expired metric, signalling to insert the metric back
+// into the parent scope.
+func (m *metricExpiryTracker) expiredAndUpdated() bool {
+	// We perform a LoadUint32 first for performance reasons
+	if atomic.LoadUint32(&m.expired) == 1 && m.scope != nil &&
```
prashantv: why do we check `scope != nil` -- it looks like it's always set in the constructors, and we never `nil` it out right?

stats.go
```diff
@@ -438,6 +568,44 @@ func newHistogramBucket(
...
+		atomic.CompareAndSwapUint32(&m.expired, 1, 0) {
+		return true
```
prashantv: everytime we return `true` here, we also call `resetUpdatedTime` in the callers. should we just call `resetUpdatedTime` here?

stats.go
```diff
@@ -438,6 +568,44 @@ func newHistogramBucket(
...
+}
+
+func (m *metricExpiryTracker) resetUpdatedTime() {
+	atomic.StoreInt64(&m.lastUpdateUnix, globalNow().Unix())
+}
+
+// checkExpiration returns whether the metric has expired and sets the expired
+// bit to one if true
+func (m *metricExpiryTracker) checkExpiration() bool {
+	if m.scope != nil && globalNow().Unix() >
```
prashantv: we're going to call `globalNow()` a ton on every report, should we do that once, and pass that in through `report`?

stats.go
```diff
@@ -438,6 +568,44 @@ func newHistogramBucket(
...
+		atomic.LoadInt64(&m.lastUpdateUnix)+m.scope.registry.expirePeriodSeconds {
+		atomic.StoreUint32(&m.expired, 1)
```
prashantv: nit: maybe a blank line before the `Store` to make it separate it from the if conditions

scope.go
prashantv: I think you can protect against that scenario with a recursive appproach. Instead of adding to a slice, you could make the current counter hold a reference to the "new" counter,

E.g.,
```go
if c.expired() {
  if newRef := c.scope.hasNew(...); newRef != nil {
    // essentially build a "linked list"
    c.newRef = newRef
  } else {
    // insert back into the scope
    c.scope.counters[...] = c
  }
}

if c.newRef != nil {
  // if we're worried about the linked list getting longer, we could also check whether `newRef` also has
  // a new ref replace `c.newRef` with `newRef.newRef`
  return c.newRef.Inc(...)
}

// actually increment this counter
```

the reporting, expiring in a scope etc all becomes a lot simpler since you won't have a list. we're building a linked list instead, but the maintenance could be simpler.

stats.go
prashantv: why do we check for `m.scope != nil`, i don't see where we `nil` out the scope, and i think all the constructos set it?

stats.go
prashantv: nit: can you put a blank line after the `{` to separate it from the multi-line condition (and can delete the one following this line)

stats.go
```diff
@@ -60,84 +60,156 @@ type counter struct {
...
+			name:           name,
+			lastUpdateUnix: globalNow().Unix(),
+		},
+	}
 }
 
 func (c *counter) Inc(v int64) {
 	atomic.AddInt64(&c.curr, v)
+
+	if c.tracker.expiredAndUpdated() {
```
prashantv: can return early here (and for the other types),
```
atomic.AddInt64(&c.curr, 1)

if !c.tracker.expiredAndUpdated() {
  return
}

// expiry handling
```

stats.go
prashantv: we always call `resetUpdatedTime` in the caller when this returns `true` right? should we just call `resetUpdatedTime` here so callers don't need to

scope.go
martin-mao: Hmm, it would make the expiry logic a little simpler, but we'll have to always traverse this linkedList to increment anything and since this is an edge case, we'd have to take a hit of checking whether this holds a ref to newRef on every Inc call, which seems like a waste for an extreme edge case (more than one element only exists when they have an old ref and inc it after the expiry period).

stats.go
martin-mao: What happens when a manager codes :P

stats.go
martin-mao: There are use cases where scope is nil (for example, the counters used within histograms). It's never niled out, but it can be constructed with no scope.

stats.go
martin-mao: Yes, we can just call here.

stats.go
martin-mao: Yeah makes sense.

scope.go
```diff
@@ -182,52 +202,190 @@ func newRootScope(opts ScopeOptions, interval time.Duration) *scope {
...
+				}
+			}
+
+			if i == 0 {
+				delete(s.counters, name)
+			} else {
+				s.counters[name] = counters[:i]
+			}
+
+			i = 0
```
prashantv: rather than resetting `i = 0`, which is easy to forget, sugget doing `var i int` inside of the for loop around `s.counters`.

same below for the other loops

scope.go
```diff
@@ -261,118 +419,233 @@ func (s *scope) reportLoopRun() {
...
+				mets += len(ss.gauges)
+				ss.gm.RUnlock()
+
+				ss.hm.RLock()
+				mets += len(ss.histograms)
+				ss.hm.RUnlock()
+
+				if mets == 0 {
+					delete(s.registry.subscopes, name)
+				}
```
prashantv: in the `else` case, should we just mark the scope as not expired?

scope.go
```diff
@@ -261,118 +419,233 @@ func (s *scope) reportLoopRun() {
 
 // reports current registry with scope status lock held
 func (s *scope) reportRegistryWithLock() {
+	scopesToRemove := map[string]struct{}{}
 	s.registry.RLock()
 	if s.reporter != nil {
-		for _, ss := range s.registry.subscopes {
-			ss.report(s.reporter)
+		for name, ss := range s.registry.subscopes {
+			if ss.report(s.reporter) {
```
prashantv: the `report` function doesn't check whether all sub-scopes are expired, assuming that's because the `registry` tracks all scopes at the root level, they're not managed like a tree right?

scope.go
```diff
@@ -437,58 +711,107 @@ func (s *scope) Capabilities() Capabilities {
...
+
+				// If there are multiple histograms created due to expiry,
+				// combine the values
+				for i := 1; i < len(histograms); i++ {
+					for k, v := range histograms[i].snapshotValues() {
+						if _, exists := values[k]; !exists {
+							values[k] = 0
+						}
+
+						values[k] = values[k] + v
```
prashantv: `values[k] += v`

same below for `durations`

scope.go
```diff
@@ -750,3 +1073,25 @@ func (s *histogramSnapshot) Values() map[float64]int64 {
 func (s *histogramSnapshot) Durations() map[time.Duration]int64 {
 	return s.durations
 }
+
+type scopeExpiryTracker struct {
+	expiredCounters   map[*counter]struct{}
+	expiredGauges     map[*gauge]struct{}
+	expiredHistograms map[*histogram]struct{}
+
+	numCounters   int
```
prashantv: why do we keep this, why not just do `len(expiredCounters)` in the `numMetrics` function, not sure it gives any performance win, since length is fast to access on a slice/map anyway. it's more likely to get out of date if we cache it separately

scope.go
martin-mao: Yeah all scopes are tracked at root level.

scope.go
martin-mao: We're tracking all counters, not just the expired ones with numCounters. We only expire the scope when all the metrics have been expired, so on that loop, len of expiredCounters will be nothing. I also need to keep track in a separate var as we read the len in a RLock.

scope.go
martin-mao: Yes, they're managed on a flat level.

mway:
Closing due to age (the next version of Tally also makes this unnecessary).

PR #98, opened 2019-04-23
Pass io.Closer in multi reporter

project0:


Project0:
Is there anything missing?

project0:
5 years later without any response, i am going to close this and will avoid OSS libs from Uber :-(

PR #100, opened 2019-05-14
Add support in `config.go` for manually disabling reporters

richardartoul:


prometheus/config.go
```diff
@@ -122,6 +130,19 @@ func (c Configuration) NewReporter(
 		opts.DefaultSummaryObjectives = values
 	}
 
+	registerer := prom.NewRegistry()
+	if !c.DisableGoCollector {
+		if err := registerer.Register(prom.NewGoCollector()); err != nil {
```
martin-mao: Nit: Why not put both checks in the same if statement?

PR #101, opened 2019-05-14
Add ability to use a non-default registry

robskillington:
This allows users to specifically enable which collectors they'd like if they are willing to use a non-default registry.

### DIFF
--- a/prometheus/config.go
+++ b/prometheus/config.go
@@ -74,8 +74,8 @@ type RegistryConfiguration struct {
 }
 
 // DefaultRegistryConfiguration returns the default registry configuration.
-// By default the default Prometheus registry is used and no collectors
-// are registered.
+// By default the default Prometheus registry already has several collectors
+// registered, such as the Go collector and the Process collector.
 func DefaultRegistryConfiguration() RegistryConfiguration {
 	return RegistryConfiguration{
 		Default: true,
--- a/prometheus/config.go
+++ b/prometheus/config.go
@@ -55,31 +55,6 @@ type Configuration struct {
 	// on the specified listen address or registering a metric with the
 	// Prometheus. By default the registerer will panic.
 	OnError string `yaml:"onError"`
-
-	// Registry specifies what registry to use and with which default
-	// collectors to use, by default the default Prometheus registry is used
-	// with whichever collectors have already been registered with the
-	// registry.
-	Registry *RegistryConfiguration `yaml:"registry"`
-}
-
-// RegistryConfiguration specifies what registry to use and with which default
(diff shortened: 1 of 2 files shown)


prometheus/config.go
```diff
@@ -53,6 +55,31 @@ type Configuration struct {
...
+// with whichever collectors have already been registered with the
+// registry.
+type RegistryConfiguration struct {
+	Default          bool `yaml:"default"`
+	GoCollector      bool `yaml:"goCollector"`
+	ProcessCollector bool `yaml:"processCollector"`
+}
+
+// DefaultRegistryConfiguration returns the default registry configuration.
+// By default the default Prometheus registry is used and no collectors
```
richardartoul: This isn’t completely accurate because the default registry registers the go/process collectors by default

prometheus/config.go
```diff
@@ -122,6 +149,22 @@ func (c Configuration) NewReporter(
...
+			}
+		}
+		if registryCfg.ProcessCollector {
+			if err := registerer.Register(prom.NewProcessCollector(os.Getpid(), "")); err != nil {
+				return nil, err
+			}
+		}
+		opts.Registerer = registerer
+	}
+
```
shreyassrivatsan: Also need to allow opening a handler against the passed in registry.. `reporter.HTTPHandler()` only registers it for the default registry, I think there is a `reporter.HandlerFor(...)` method..

prometheus/config.go
robskillington: Sure, but I don't actually register further collectors - which you could do if you enable them.

I'll make it more clear though, as yes the default registry comes with some collectors.

prometheus/config.go
robskillington: Good call, fixed with https://github.com/uber-go/tally/pull/80 and the changes I just made to that code now to default to that registerer.

prometheus/config.go
```diff
@@ -53,6 +55,31 @@ type Configuration struct {
...
+	// with whichever collectors have already been registered with the
+	// registry.
+	Registry *RegistryConfiguration `yaml:"registry"`
+}
+
+// RegistryConfiguration specifies what registry to use and with which default
+// collectors to use, by default the default Prometheus registry is used
+// with whichever collectors have already been registered with the
+// registry.
+type RegistryConfiguration struct {
```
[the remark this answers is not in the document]
robskillington: True, that's plausible.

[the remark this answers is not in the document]
robskillington: In fact, much much better - I like that. 

Will change to that suggested pattern.

PR #105, opened 2019-06-20
Proposal to create a NewRegistry by default

dlmather:
Currently the default behavior on creating a NewReporter is to use the prometheus DefaultReporter, which can result in, for instance, difficult to track down panics due to duplicate metric registry when spinning up two different reporters in separate test cases. I would like to propose either creating a NewRegistry by default, or at least documenting the default behavior of this method. This is just a suggestion and I'm open to learning the reason that a global is used by default.


zmt:
Excellent find of your test failure root cause. I don't have the context behind the decision to use the `prometheus.DefaultRegisterer`. I suspect this could break some consumers or making this change might require a little more, but I don't know this codebase well.

PR #112, opened 2019-11-07
Simplify mutex use

### LINKED ISSUE #111: Tally deadlocks if the reporter panics.
If AllocateCounter() below fails, , s.cm won't be unlocked and further calls to scope.Tagged() will deadlock.

```
func (s *scope) Counter(name string) Counter {
  name = s.sanitizer.Name(name)
  s.cm.RLock()
  val, ok := s.counters[name]
  s.cm.RUnlock()
  if !ok {
    s.cm.Lock()
    val, ok = s.counters[name]
    if !ok {
      var cachedCounter CachedCount
      if s.cachedReporter != nil {
        cachedCounter = s.cachedReporter.AllocateCounter(                                                                                                                                                                        
          s.fullyQualifiedName(name), s.tags,
        )
      }
      val = newCounter(cachedCounter)
      s.counters[name] = val
    }
    s.cm.Unlock()
  }
  return val
}
```

This is occurring with the prometheus reporter.

mway:
Fixes #111.

There's no real reason to attempt to hyper-optimize mutexes in this way; the cost of `defer` is negligible (for example, average counter allocation increased from 515ns to 580ns as part of this change) and the safety of deferring unlocks would be worth an even more substantial penalty.

Wherever this minor performance degradation would make a difference, metrics should be being cached anyway.

### DIFF
--- a/scope.go
+++ b/scope.go
@@ -245,19 +245,20 @@ func (s *scope) reportLoopRun() {
 	// Need to hold a status lock to ensure not to report
 	// and flush after a close
 	s.status.RLock()
+	defer s.status.RUnlock()
+
 	if s.status.closed {
-		s.status.RUnlock()
 		return
 	}
 
 	s.reportRegistryWithLock()
-
-	s.status.RUnlock()
 }
 
 // reports current registry with scope status lock held
 func (s *scope) reportRegistryWithLock() {
 	s.registry.RLock()
+	defer s.registry.RUnlock()
+
 	if s.reporter != nil {
 		for _, ss := range s.registry.subscopes {
 			ss.report(s.reporter)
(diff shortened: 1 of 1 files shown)


prometheus/reporter.go
```diff
@@ -291,6 +291,18 @@ func NewReporter(opts Options) Reporter {
 	}
 	if opts.OnRegisterError == nil {
 		opts.OnRegisterError = func(err error) {
+			// n.b. Because our forked Prometheus client does not actually emit
```
jeffbean: this seems to maybe warrant its own PR. lets not overload this diff with different concerns. Even though its to allow us to test i think since this is changing core behavior it should be its own diff.

mway: I agree, but behavior isn't changing here - just the error that's being panicked. Users who provide their own `OnRegisterError` will see no difference as a result of this change.

The only difference that updating our Prom fork would do is allowing us to type switch on the error, and that comes at an unknown (but potentially high) cost of upgrading dependents' use of the Prom client, which I don't think is quite as valuable here. They're two distinct concerns, but one does not necessarily follow from the other.

mway: (The same holds for would-be recoverers of this panic - this is a builtin error, so they will receive the same underlying error type via any `recover()` calls, modulo the additional (arguably more helpful) error message.

jeffbean: ok, seems your case is reasonable.

prometheus/reporter_test.go
```diff
@@ -358,6 +358,29 @@ func TestOnRegisterError(t *testing.T) {
...
+	})
+
+	r.AllocateCounter("foo", map[string]string{"foo": "bar"})
+	r.AllocateCounter("foo", map[string]string{"foo": "baz"}) // ok; same label names
+	r.AllocateCounter("foo", map[string]string{"bar": "qux"}) // errors; different label names
+	r.AllocateCounter("foo", map[string]string{
+		"bar":  "qux",
+		"zork": "derp",
+	}) // errors; different label names
+	r.AllocateCounter("foo", nil) // errors; different label names
```
jeffbean: it's not obvious from `different label names ` why this is an error.

can we assert on an error string directly here?

mway: sure! added some docs as well.

prometheus/reporter.go
prashantv: I'd still vote for this to go in a separate PR, given it's not related to mutex changes. It seems like a very independent change (definitely a good change, just not related to mutexes)

glide.yaml
```diff
@@ -17,6 +17,8 @@ import:
   version: ^0.8.1
 - package: go.uber.org/atomic
   version: ^1
+- package: github.com/pkg/errors
```
prashantv: can we move this out to a separate change, not really related to mutexes

scope.go
```diff
@@ -269,16 +270,16 @@ func (s *scope) reportRegistryWithLock() {
...
 	}
-	s.registry.RUnlock()
 }
 
 func (s *scope) Counter(name string) Counter {
 	name = s.sanitizer.Name(name)
-	s.cm.RLock()
+
+	s.cm.Lock()
+	defer s.cm.Unlock()
```
prashantv: This could have large performance implications, since an `RLock` allows multiple readers, while `Lock` only allows one reader.

I think it's common for callers to not hold a reference to a `Counter`, and to rely on this cache, and changing to a lock can significantly increase contention.

I'd suggest changing this to:
```go
func (s *scope) getCounter(name string) (Counter, bool) {
  s.cm.RLock()
  defer s.cm.RUnlock()

  val, ok := s.counters[name]
  return val, ok
}

func (s *scope) Counter(name string) Counter {
  if c, ok := s.getCounter(name); ok {
    return c
  }

  s.cm.Lock()
  defer s.cm.Unlock()

   if val,  ok := s.counters[name]; ok {
    return val
  }
  
  var cachedCounter CachedCount
  val = newCounter(cachedCounter)
  s.counters[name] = val
  return val
}
```

as a bonus, it significantly reduces the amount of nesting,
https://github.com/uber-go/guide/blob/master/style.md#reduce-nesting

same is true for `Gauge`, `Timer`, etc below

mway: I explained in the description that benchmarks are negligibly impacted. Also, pedantically, `getX` is not a recommended Go naming convention:

> it's neither idiomatic nor necessary to put Get into the getter's name

(ref: [Effective Go](https://golang.org/doc/effective_go.html#Getters))

I'll make the change recommended here to minimize internal version churn, but let's put any additional changes into subsequent PRs. Sound fair?

glide.yaml
mway: I think this is a fair tradeoff for not implementing (however temporary) ad hoc error wrapping, e.g.:

```go
err = fmt.Errorf("something bad happened: %v", err)
```

is not conventionally used so much as:

```go
err = errors.WithMessage(err, "something bad happened")
```

(same with `errors.Wrap`, etc)

scope.go
prashantv: Is there benchmarks that test the contention caused by concurrency? I think we'll only see impact from benchmarks in that scenario.

Agree that lack of nesting can be done separately, just want to make sure we keep using `RLock` here, thanks

glide.yaml
prashantv: Agree, but I think that whole change (changing the error message) should be done separately since it isn't really related to the mutex simplification.

scope.go
```diff
@@ -269,108 +270,155 @@ func (s *scope) reportRegistryWithLock() {
 		}
 		s.cachedReporter.Flush()
 	}
-	s.registry.RUnlock()
 }
 
 func (s *scope) Counter(name string) Counter {
+	if c, ok := s.counter(name); ok {
```
prashantv: i think the sanitizer.Name needs to be called before we do this check here? maybe the argument name in `counter` should be called `sanitizedName` to make it clearer as well?

mway: yeah, nice catch. thanks!

scope.go
```diff
@@ -387,16 +431,13 @@ func (s *scope) subscope(prefix string, immutableTags map[string]string) Scope {
 	immutableTags = mergeRightTags(s.tags, immutableTags)
 	key := scopeRegistryKey(prefix, immutableTags)
 
-	s.registry.RLock()
+	s.registry.Lock()
```
prashantv: we currently have `existing, ok := s.registry.subscopes[key]` twice below, which is probably a remnant of checking under rlock vs lock.

To avoid any performance difference, suggest doing the same thing as the counters etc by having the rlock check first in a separate function.

types.go
```diff
@@ -28,6 +28,11 @@ import (
 
 // Scope is a namespace wrapper around a stats reporter, ensuring that
 // all emitted values have a given prefix or set of tags.
+//
```
prashantv: this probably belongs in the separate prob-related PR as well since it's not related to the mutexes

scope.go
mway: done!

types.go
mway: 👍

PR #119, opened 2020-01-09
m3: Remove atomic from TCalcTransport

prashantv:
The TCalcTransport is used by a single caller at a time, with a
mutex used to protect access:
https://github.com/uber-go/tally/blob/5431c04afc0bad76ce91ca3d549abb469abd2a10/m3/reporter.go#L400

The thrift generated code will only call `Write` serially, so there's no
need to use atomics here.

Removal of atomics gives a small performance improvement (10-15%):
```
> benchcmp -best old new
benchmark                 old ns/op     new ns/op     delta
BenchmarkCalulateSize     538           453           -15.80%
```

### DIFF
--- a/m3/customtransports/m3_calc_transport.go
+++ b/m3/customtransports/m3_calc_transport.go
@@ -21,7 +21,7 @@
 package customtransport
 
 // TCalcTransport is a thrift TTransport that is used to calculate how many
-// bytes are used when writing a thrift element. It is thread-safe
+// bytes are used when writing a thrift element.
 type TCalcTransport struct {
 	count int32
 }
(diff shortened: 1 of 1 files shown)


m3/customtransports/m3_calc_transport.go
```diff
@@ -20,10 +20,6 @@
 
 package customtransport
 
-import (
-	"sync/atomic"
-)
-
```
prateek: should probably remove the "It is thread-safe" comment on line 28 below.

prashantv: Good catch, removed.

PR #121, opened 2020-01-14
Optimize Scope.cacheReport

prateek:
Optimize the cacheReport path for all Scopes.

Results for the added benchmark (master v HEAD):
```
benchmark                                  old ns/op      new ns/op     delta
BenchmarkScopeReporting/size1-4            79.6           42.0          -47.24%
BenchmarkScopeReporting/size10-4           201            62.3          -69.00%
BenchmarkScopeReporting/size100-4          1666           282           -83.07%
BenchmarkScopeReporting/size1000-4         16439          2547          -84.51%
BenchmarkScopeReporting/size10000-4        167362         26812         -83.98%
BenchmarkScopeReporting/size100000-4       4239618        328205        -92.26%
BenchmarkScopeReporting/size1000000-4      69189550       5549203       -91.98%
BenchmarkScopeReporting/size10000000-4     2205450363     35460112      -98.39%
```

Map iteration << Slice iteration

### DIFF
--- a/scope.go
+++ b/scope.go
@@ -27,7 +27,7 @@ import (
 )
 
 const (
-	defaultInitialSliceSize = 16
+	_defaultInitialSliceSize = 16
 )
 
 var (
@@ -79,10 +79,11 @@ type scope struct {
 	countersSlice   []*counter
 	gauges          map[string]*gauge
 	gaugesSlice     []*gauge
-	timers          map[string]*timer
-	timersSlice     []*timer
 	histograms      map[string]*histogram
 	histogramsSlice []*histogram
+	timers          map[string]*timer
+	// nb: deliberately skipping timersSlice as we report timers immediately,
+	// no buffering is involved.
 }
 
 type scopeStatus struct {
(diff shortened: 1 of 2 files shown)


scope.go
```diff
@@ -72,10 +76,14 @@ type scope struct {
...
-	counters   map[string]*counter
-	gauges     map[string]*gauge
-	timers     map[string]*timer
-	histograms map[string]*histogram
+	counters        map[string]*counter
+	countersSlice   []*counter
+	gauges          map[string]*gauge
+	gaugesSlice     []*gauge
+	timers          map[string]*timer
+	timersSlice     []*timer
```
vdarulis: Timers aren't used in `CacheReport`, as they report directly.

scope_benchmark_test.go
```diff
@@ -121,3 +122,82 @@ func BenchmarkHistogramExisting(b *testing.B) {
...
+	}
+	b.ResetTimer()
+
+	for n := 0; n < b.N; n++ {
+		s.cachedReport()
+	}
+}
+
+func BenchmarkScopeReporting(b *testing.B) {
+	for i := 1; i <= 10000000; i *= 10 {
```
vdarulis: 10M is a bit overkill/unrealistic, limit to 1M so it completes faster?

scope.go
```diff
@@ -72,10 +76,14 @@ type scope struct {
 	tm sync.RWMutex
 	hm sync.RWMutex
```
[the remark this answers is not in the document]
mway: Let's do this in a followup - I have a parallel reporting change that this may be beneficial for (and can be proven with benchmarks, TBD).

scope.go
```diff
@@ -72,10 +76,14 @@ type scope struct {
...
 
-	counters   map[string]*counter
-	gauges     map[string]*gauge
-	timers     map[string]*timer
-	histograms map[string]*histogram
+	counters        map[string]*counter
+	countersSlice   []*counter
```
mway: nit: Can we find a better name than `fooSlice`?

scope.go
```diff
@@ -27,6 +27,10 @@ import (
 	"time"
 )
 
+const (
+	defaultInitialSliceSize = 16
```
mway: style: underscore prefix

scope.go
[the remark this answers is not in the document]
prateek: yeah out of scope for this change ..

scope.go
prateek: sure

scope.go
prateek: removed and mentioned with a comment

scope_benchmark_test.go
prateek: sure

scope.go
prateek: thought about a few different choices:
- `fooList`, `fooCollection`, etc: all seem worse than `fooSlice`.
- `fooIteration` or something along those lines to indicate it's for iteration but that seems even more artificial

need this for something so going to land as is. if you have a better suggestion, i'll open a followup.

PR #127, opened 2020-03-05
Get rid of glide and use Go modules

vdarulis:


Makefile
```diff
@@ -1,61 +1,62 @@
...
+	if [ -s lint.log ]; then \
+		echo lint or fmt failed! errors:  ;\
+		cat lint.log ;\
+		exit 1 ;\
+	fi
 
 .PHONY: test
 test:
-	go test -race -v $(PKGS)
+	$(GO) test -timeout 1m -race -v ./...
```
mway: here and in `cover`, you've already expanded `./...` in `$(PKGS)` so you might as well use that, otherwise you're just doing another `go list` under the hood

Makefile
```diff
@@ -1,61 +1,62 @@
...
-	glide --version || go get -u -f github.com/Masterminds/glide
-	glide install
 	@echo "Installing test dependencies..."
-	go install ./vendor/github.com/axw/gocov/gocov
-	go install ./vendor/github.com/mattn/goveralls
+	$(GO) get github.com/axw/gocov/gocov
+	$(GO) get github.com/mattn/goveralls
 	@echo "Installing golint..."
-	go install ./vendor/github.com/golang/lint/golint
+	$(GO) get -u golang.org/x/lint/golint
```
mway: nit: don't need/want `-u` here unless we're going to use it for all of them; don't want some packages being updated while others aren't

lukashes:
Hi! Looks like this PR doesn't meet the rules of v3 gomod packages

vdarulis:
@lukashes thanks, will take a look - haven't cut a release yet, but definitely might have missed something in testing go.mod <-> go.mod and not legacy package managers.

PR #133, opened 2020-03-27
subscoping: Defer tag merging

abhinav:
**Commits should be reviewed individually.**

This improves the performance of creating subscopes with new tags by
deferring merged map creation.

We currently create the merged map of tags first, and then calculate the
key which we use to determine if we have a cached subscope.

With this change, we calculate the key with slightly more complex logic
that inspects the unmerged maps and avoids paying the cost of merging
the maps if not needed.

To make this change, we first refactor scopeRegistry so that Scope
accesses it like a separate object.

---

I ran the benchmarks with `-count 3`.

**Before**

```
BenchmarkScopeTaggedCachedSubscopes-4             486406              2334 ns/op            1152 B/op          9 allocs/op
BenchmarkScopeTaggedCachedSubscopes-4             445890              2567 ns/op            1152 B/op          9 allocs/op
BenchmarkScopeTaggedCachedSubscopes-4             423003              2721 ns/op            1152 B/op          9 allocs/op
BenchmarkScopeTaggedNoCachedSubscopes-4           189488              6149 ns/op            2203 B/op         17 allocs/op
BenchmarkScopeTaggedNoCachedSubscopes-4           273436              6063 ns/op            2235 B/op         17 allocs/op
BenchmarkScopeTaggedNoCachedSubscopes-4           278360              5049 ns/op            2233 B/op         17 allocs/op
```

**After**

```
BenchmarkScopeTaggedCachedSubscopes-4             652670              1791 ns/op             832 B/op          8 allocs/op
BenchmarkScopeTaggedCachedSubscopes-4             669008              1953 ns/op             832 B/op          8 allocs/op
BenchmarkScopeTaggedCachedSubscopes-4             591230              1817 ns/op             832 B/op          8 allocs/op
BenchmarkScopeTaggedNoCachedSubscopes-4           234835              5876 ns/op            2269 B/op         18 allocs/op
BenchmarkScopeTaggedNoCachedSubscopes-4           315829              5058 ns/op            2237 B/op         18 allocs/op
BenchmarkScopeTaggedNoCachedSubscopes-4           243750              4993 ns/op            2264 B/op         18 allocs/op
```

On average,

- Cached hits are ~27% *faster*: ~687 ns/op
- Cache misses are ~7% *faster*: ~444 ns/op

### DIFF
--- a/key_gen.go
+++ b/key_gen.go
@@ -54,11 +54,20 @@ func KeyForPrefixedStringMap(
 	prefix string,
 	stringMap map[string]string,
 ) string {
-	keys := keyGenPool.stringsPool.Get().([]string)
-	for k := range stringMap {
-		keys = append(keys, k)
-	}
+	return keyForPrefixedStringMaps(prefix, stringMap)
+}
 
+// keyForPrefixedStringMaps generates a unique key for a prefix and a series
+// of maps containing tags.
+//
+// If a key occurs in multiple maps, keys on the right take precedence.
+func keyForPrefixedStringMaps(prefix string, maps ...map[string]string) string {
+	keys := keyGenPool.stringsPool.Get().([]string)
+	for _, m := range maps {
+		for k := range m {
+			keys = append(keys, k)
+		}
+	}
 	sort.Strings(keys)
 
 	buf := keyGenPool.bufferPool.Get().(*bytes.Buffer)
@@ -68,13 +77,27 @@ func KeyForPrefixedStringMap(
 		buf.WriteByte(prefixSplitter)
 	}
 
-	sortedKeysLen := len(stringMap)
-	for i := 0; i < sortedKeysLen; i++ {
-		buf.WriteString(keys[i])
-		buf.WriteByte(keyNameSplitter)
-		buf.WriteString(stringMap[keys[i]])
-		if i != sortedKeysLen-1 {
+	var lastKey string // last key written to the buffer
+	for _, k := range keys {
+		if len(lastKey) > 0 {
+			if k == lastKey {
+				// Already wrote this key.
+				continue
+			}
 			buf.WriteByte(keyPairSplitter)
+		}
+		lastKey = k
+
+		buf.WriteString(k)
+		buf.WriteByte(keyNameSplitter)
+
+		// Find and write the value for this key. Rightmost map takes
+		// precedence.
+		for j := len(maps) - 1; j >= 0; j-- {
+			if v, ok := maps[j][k]; ok {
+				buf.WriteString(v)
+				break
+			}
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


key_gen.go
```diff
@@ -54,12 +54,22 @@ func KeyForPrefixedStringMap(
...
 	stringMap map[string]string,
 ) string {
+	return keyForPrefixedStringMaps(prefix, stringMap)
+}
+
+// keyForPrefixedStringMaps generates a unique key for a prefix and a series
+// of maps containing tags.
+//
+// If a key occurs in multiple maps, keys on the right take precedence.
+func keyForPrefixedStringMaps(prefix string, maps ...map[string]string) string {
```
vdarulis: if you wanted to take this a step further, it would be even more performant to keep slices of tags (key/value pairs) and then just merging those with user-provided tag map on scope creation.

abhinav: Yeah, there's more room for optimization on this code path. I may explore it more.
This was just a low hanging fruit we ran into during a conversation on Friday.

key_gen.go
```diff
@@ -68,12 +78,20 @@ func KeyForPrefixedStringMap(
...
-	for i := 0; i < sortedKeysLen; i++ {
-		buf.WriteString(keys[i])
+	for i, k := range keys {
+		buf.WriteString(k)
 		buf.WriteByte(keyNameSplitter)
-		buf.WriteString(stringMap[keys[i]])
-		if i != sortedKeysLen-1 {
+
+		// Find and write the value for this key. Rightmost map takes
+		// precdence.
```
prashantv: nit: typo, precedence

key_gen.go
```diff
@@ -68,12 +78,20 @@ func KeyForPrefixedStringMap(
...
+		for j := len(maps) - 1; j >= 0; j-- {
```
prashantv: (idea for follow up): if instead of storing `[]string`, if we instead had `[]struct{key string, index int}`, dedupe would keep the highest precedence map and we could avoid a loop here.

key_gen.go
```diff
@@ -83,6 +101,24 @@ func KeyForPrefixedStringMap(
...
+func dedupeSortedStrings(in []string) (out []string) {
+	if len(in) == 0 {
+		return in
+	}
+
+	prev := in[0]
+	out = in[:1] // zero-alloc filtering
+	for _, v := range in[1:] {
+		if prev == v {
+			continue
```
prashantv: separate method is good for breaking up the logic of the above function, but this could probably be inlined into the for loop around keys. that way we avoid iterating over the same slice twice, and it's not much additional code (just keep `prev` and the check against `prev`)

abhinav: Good call. Switching.

PR #134, opened 2020-03-31
Revert "Get rid of glide and use Go modules (#127)"

vdarulis:
This reverts commit 798e6594bd9466947d80e8c18252d45e6b37d398.

We're in a semver pickle. If we migrate to go.mod - we'll need our users to use the /v3 import path.


robsonpeixoto:
Are there plans to use `go.mod` again ?

PR #135, opened 2020-04-16
Allow Prometheus reporter listen on unix domain socket address

biefy:
This PR will allow prometheus reporter listen on unix domain socket address. This is very useful when running under a situation when port conflicts occur often.

Currently, if specify a UDS address like `unix:/tmp/tally-metrics.sock` as listening address, reporter would fail to start as `http.ListenAndServe()` used by the reporter only support `tcp` network.

### DIFF
--- a/prometheus/config.go
+++ b/prometheus/config.go
@@ -25,7 +25,6 @@ import (
 	"log"
 	"net"
 	"net/http"
-	"net/url"
 	"os"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


biefy:
Hi @prateek  @robskillingtonm @mway ,  could you please help to take a look this PR?

prometheus/config.go
```diff
@@ -148,7 +153,20 @@ func (c Configuration) NewReporter(
...
+			nwk := c.ListenNetwork
+			if nwk == "" {
+				nwk = "tcp"
+			}
+
+			lsn, err := net.Listen(nwk, addr)
+			if err != nil {
+				opts.OnRegisterError(err)
+				return
+			}
```
vdarulis, line 165: can you keep the long-form names (`network`, `listener`)

biefy: @vdarulis thank you. updated.

PR #138, opened 2020-05-20
Removing m3db forked packages

nikunjy:
Tally is using bunch of m3db forked packages. These forks don't provide any use and they prevent from using the rest of the Prometheus library in conjunction with tally. 
For ex I should be able to do this:
```
 reg := prometheus.NewRegistry()
grpcMetrics := grpc_prometheus.NewServerMetrics()
reg.MustRegister(grpcMetrics)
var registerer prometheus.Registerer = reg
r := promreporter.NewReporter(promreporter.Options{
  Registerer: reg,
})
```

### DIFF
--- a/prometheus/config.go
+++ b/prometheus/config.go
@@ -28,7 +28,7 @@ import (
 	"os"
 	"strings"
 
-	prom "github.com/m3db/prometheus_client_golang/prometheus"
+	prom "github.com/prometheus/client_golang/prometheus"
 )
 
 // Configuration is a configuration for a Prometheus reporter.
--- a/prometheus/reporter.go
+++ b/prometheus/reporter.go
@@ -26,9 +26,9 @@ import (
 	"sync"
 	"time"
 
-	prom "github.com/m3db/prometheus_client_golang/prometheus"
-	"github.com/m3db/prometheus_client_golang/prometheus/promhttp"
 	"github.com/pkg/errors"
+	prom "github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/uber-go/tally"
 )
 
@@ -163,8 +163,8 @@ type cachedMetric struct {
 	counter     prom.Counter
 	gauge       prom.Gauge
 	reportTimer func(d time.Duration)
-	histogram   prom.Histogram
-	summary     prom.Summary
+	histogram   prom.Observer
+	summary     prom.Observer
 }
 
 func (m *cachedMetric) ReportCount(value int64) {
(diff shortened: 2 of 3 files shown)


nikunjy:
what do you think @robskillington ?

mway:
Hey @nikunjy, thanks for the PR! There are internal dependency graphs that need to be considered as well, so we'll need someone in Uber to merge this once we're able to validate what we do/don't need from a forking/pinning/etc perspective.

No ETA on this yet, but we're generally interested in un-unicorning things as much as possible. Is there a specific urgency on your end?

vdarulis:
FYI we cannot upgrade to go modules, I tried and reverted - it is a major breaking change as our users have to migrate to a new import path.

nikunjy:
@mway and @vdarulis thanks for taking a look. First of all, tally is a great library :) 
I forked a version at brex so it is not urgent, was just looking to possibly make this one better. 

Have you considered using the new aliases introduced in Go 1.13 It might make it easy for you to make a backwards compatible change . 

@vdarulis  why would users have to upgrade to a new import path ? I don't follow

vdarulis:
>  @vdarulis why would users have to upgrade to a new import path ? I don't follow
Not all of our users have migrated to go modules. If Tally is migrated to go modules, to use newer versions users will have to change their import path to `github.com/uber-go/tally/v3` (or `/v4`, since technically go-mod migration is a big breaking change). Also, if we update the path, all of our users still on `dep` will be unable to use the new versions properly.

While I'd love to switch, the cons are too many (for now) for a small improvement in devexp.

nikunjy:
@vdarulis  What bout just removing m3db forks without the go mod update. I just added the go mod commit because I set my local setup like that.

robskillington:
Hey good to see this conversation happening and TY for bringing this up @nikunjy .

The main reason it was forked in the first place was that `github.com/prometheus/client_golang` client library was previously using below version v1 (i.e. v0.x) and was making breaking changes constantly.

This meant tally could not be used inside Uber because a lot of projects depended on code that referenced some different version of the Prometheus client library than what `tally` was vendored against.

An easier fix would be to remove the Prometheus reporter altogether from tally and implement it in a separate repository that has a version for every Prometheus client library version that is published, that way you could always use the relevant Prometheus client library (per what version of Prometheus client library you are using internally).

The alternative is to start depending on Prometheus client library directly since now it is 1.x. This still has the problem that @mway and @vdarulis have called out, in that if an end user application already depends on either 0.x Prometheus client library or 1.x then until they are using `go.mod` they're not going to be able to use `tally` if they have a dependency (or directly depends) on a different version of Prometheus client library (i.e. if they transitively depend on 0.x and tally depends on 1.x or the inversion of that).

Issue #141, opened 2020-06-02
scope.Snapshot() clears Histogram values


jkanywhere:
### Steps to Reproduce:

Instantiate a testScope.
Emit some metrics.

Call `Snapshot()` multiple times on a `testScope`.

As demonstrated by running existing test in a loop: https://github.com/uber-go/tally/compare/master...jkanywhere:jkline-histogram-snapshot-broken 

### Expected Result

Value of metrics is the same after each call to `Snapshot()` because no new metrics have been emitted.

### Actual Result

The expected result holds for counters, gauges, and times.
But, the value stored in Histograms is cleared after each call to `Snapshot()`.

### Analysis

The value stored in Histograms is cleared after each call to `Snapshot()`. Applies to both Value buckets and Duration buckets.

Am I misunderstanding the intended use of `Snapshot()`? I've been assuming it is safe to call multiple times during the same test.

A histogram is a map of float or duration to `counter`.
`(*histogram).snapshotValues()` and `(*histogram).snapshotDurations()` each call `(*counter).value()` for each bucket.

Calling `(*counter).value()` changes the value stored: https://github.com/uber-go/tally/blob/master/stats.go#L73

Should probably call `(*counter).snapshot()` instead, as is done when snapshotting counters themselves.

mway:
this has been released in [v3.3.17](https://github.com/uber-go/tally/releases/tag/v3.3.17), feel free to update pins as necessary.

thanks for fixing this!

PR #142, opened 2020-06-03
Don't change Histogram values during Snapshot

jkanywhere:
Make it idempotent to call `(*scope).Snapshot` multiple times during the same test case by using `(*counter).snapshot` instead of `(*counter).value` to capture the value of a Histogram during `Snapshot`. `value` changes the underlying value, `snapshot` does not.

https://github.com/uber-go/tally/issues/141

### DIFF
--- a/stats.go
+++ b/stats.go
@@ -382,7 +382,7 @@ func (h *histogram) snapshotValues() map[float64]int64 {
 
 	vals := make(map[float64]int64, len(h.buckets))
 	for i := range h.buckets {
-		vals[h.buckets[i].valueUpperBound] = h.buckets[i].samples.value()
+		vals[h.buckets[i].valueUpperBound] = h.buckets[i].samples.snapshot()
 	}
 
 	return vals
@@ -395,7 +395,7 @@ func (h *histogram) snapshotDurations() map[time.Duration]int64 {
 
 	durations := make(map[time.Duration]int64, len(h.buckets))
 	for i := range h.buckets {
-		durations[h.buckets[i].durationUpperBound] = h.buckets[i].samples.value()
+		durations[h.buckets[i].durationUpperBound] = h.buckets[i].samples.snapshot()
 	}
 
 	return durations
--- a/scope_test.go
+++ b/scope_test.go
@@ -21,6 +21,7 @@
 package tally
 
 import (
+	"fmt"
 	"math"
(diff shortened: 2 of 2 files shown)


jkanywhere:
Test failure is only for go tip, and seems unrelated:
https://travis-ci.org/github/uber-go/tally/jobs/694067010
```
FAIL	github.com/uber-go/tally/thirdparty/github.com/apache/thrift/lib/go/thrift [build failed]
```

stats.go
```diff
@@ -382,7 +382,7 @@ func (h *histogram) snapshotValues() map[float64]int64 {
 
 	vals := make(map[float64]int64, len(h.buckets))
 	for i := range h.buckets {
-		vals[h.buckets[i].valueUpperBound] = h.buckets[i].samples.value()
+		vals[h.buckets[i].valueUpperBound] = h.buckets[i].samples.snapshot()
```
zmt, line 385: This maps to https://sourcegraph.uberinternal.com/github.com/uber-go/tally/-/blob/stats.go#L102 since `h.buckets` here are `[]histogramBucket` and `histogramBucket.samples` are `*counter`. This looks correct vs. prior use of `value()`(https://sourcegraph.uberinternal.com/github.com/uber-go/tally/-/blob/stats.go#L73).

stats.go
```diff
@@ -395,7 +395,7 @@ func (h *histogram) snapshotDurations() map[time.Duration]int64 {
 
 	durations := make(map[time.Duration]int64, len(h.buckets))
 	for i := range h.buckets {
-		durations[h.buckets[i].durationUpperBound] = h.buckets[i].samples.value()
+		durations[h.buckets[i].durationUpperBound] = h.buckets[i].samples.snapshot()
```
zmt, line 398: Same here.

stats.go
mway: yep, that's right. nice catch, thanks for fixing!

also: please use sourcegraph.com URLs for OSS projects.

mway:
will merge this in once travis is done running again; may be some issues with Go tip, but we can sort those out separately if so.

stats.go
zmt: > also: please use sourcegraph.com URLs for OSS projects.

Derp. Corrected.

PR #144, opened 2020-07-18
use the testify functions for testing errors

trashhalo:
[NoError](https://godoc.org/github.com/stretchr/testify/require#NoError) is for asserting no error was return
[Error](https://godoc.org/github.com/stretchr/testify/require#Error) is for asserting an error was returned


mway (approved):
Thank you!

PR #146, opened 2021-02-06
Make histogram allocations more efficient, general cleanup

mway:
Histogram types are mutually exclusive, so there's no need to allocate storage for both types regardless of which type of bucket is provided.

Also includes some miscellaneous tweaks/cleanup.

At 100k histograms, this reduces their memory footprint by 45-65%, with diminishing returns of ~2% every 100k thereafter.

### DIFF
--- a/m3/reporter.go
+++ b/m3/reporter.go
@@ -21,15 +21,16 @@
 package m3
 
 import (
-	"errors"
 	"fmt"
 	"io"
 	"math"
 	"os"
+	"sort"
 	"strconv"
 	"sync"
 	"time"
 
+	"github.com/pkg/errors"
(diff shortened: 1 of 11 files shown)


m3/reporter.go
```diff
@@ -288,45 +323,57 @@ func (r *reporter) AllocateHistogram(
 	buckets tally.Buckets,
 ) tally.CachedHistogram {
```
mway, line 324: Note: as a followup, I'll probably refactor histograms to share the same tag storage and store their separate bucket names (instead of separate tag maps), and resolve the tags at reporting time (potentially via map pooling) instead of allocating like crazy.

vdarulis (approved):

PR #148, opened 2021-02-06
Tags support for statsd reporter

jredl-va:
The cactus library has recently released tags support which has been missing from the statsd reporter: https://github.com/cactus/go-statsd-client/releases/tag/v5.0.0

- Update dependencies and use the new tags interface. 
- Go modules support 

The go modules part of this feels like it should be a separate unit of work when comparing it against the work done for zap: https://github.com/uber-go/zap/pull/751/files


jredl-va:
Actually...looking at this again, I can just provide my own implementation to tally using v5 of cactus.

PR #149, opened 2021-02-22
Thrift rework for allocations, more histogram improvements

mway:
**This is a breaking change (wire compatibility) based on top of #146.**

tl;dr This PR introduces a new version for the Tally<>M3 Thrift API and includes several optimizations to avoid unnecessary memory usage, particularly when using histograms.

- Adds a new Thrift output for M3 reporters
- Restructures Thrift imports slightly (`m3/thrift` -> `m3/thrift/{v1,v2}`)
- Removes all unnecessary Thrift pointers
- Changes some Thrift message semantics (e.g. no more pointer-indexed maps, so `map[*m3.MetricTag]bool` is now `[]m3.MetricTag`)
- Shares all common histogram storage between allocated histograms
- Adds a histogram bucket cache to avoid recomputing internal bucket storage for the same set of buckets
- Builds CI against 1.15.x and 1.16.x, and removes tip
- Other miscellaneous optimizations and cleanup


m3/thrift/v2/v2.thrift
```diff
@@ -0,0 +1,34 @@
...
+    COUNTER = 1
+    GAUGE = 2
+    TIMER = 3
+}
+
+struct MetricValue {
+    1: required MetricType metricType
+    2: required i64 count
+    3: required double gauge
+    4: required i64 timer
```
vdarulis, line 12: wait, we're sending each timer sample as a Metric payload, even back in V1 protocol?! can we at least make this a list in v2 thrift

vdarulis (approved):
lgtm with some comments

m3/reporter.go
```diff
@@ -288,45 +331,88 @@ func (r *reporter) AllocateHistogram(
...
+					valueUpperBound:    pair.UpperBoundValue(),
+					durationUpperBound: pair.UpperBoundDuration(),
+					metric:             counter,
+				},
+			)
+		} else {
+			var (
+				counter = r.allocateCounter(name, htags)
+				bucket  = r.valueBucketString(pair.LowerBoundValue()) +
+					"-" + r.valueBucketString(pair.UpperBoundValue())
```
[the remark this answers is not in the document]
mway: agreed. golfed almost all of it away now.

m3/thrift/v2/v2.thrift
mway: Discussed offline; will do this as a followup.

vdarulis (approved):

internal/identity/accumulator.go
```diff
@@ -0,0 +1,89 @@
...
+func (a Accumulator) AddUint64(u64 uint64) Accumulator {
+	return a + Accumulator(u64*_hashFold)
+}
+
+// AddUint64s serially folds each of u64s into the accumulator.
+//go:nosplit
+func (a Accumulator) AddUint64s(u64s ...uint64) Accumulator {
+	for _, u64 := range u64s {
+		a += Accumulator(u64 * _hashFold)
+	}
```
vdarulis, line 80: maybe you could replace for with a goto to make this inline-able before go 1.16? if you're looking for extra edge

internal/identity/accumulator.go
```diff
@@ -0,0 +1,89 @@
...
+	_hashSeed uint64 = 23
+	_hashFold uint64 = 31
+)
+
+// Accumulator is a commutative folding accumulator.
+type Accumulator uint64
+
+// NewAccumulator creates a new Accumulator with a default seed value.
+//
+// n.b. Here and elsewhere, we use nosplit to avoid stack size checks, which
```
vdarulis, line 37: thanks!

vdarulis:
On another thought, are you sure v1/v2 can coexist? 
We use a custom transport, so service name does not factor in anywhere, just the RPC method - which is the same across both v1/v2, so there's no way to distinguish them if we want to support both on the same port.

internal/identity/accumulator.go
prashantv: I'm a little cautious when using the Go pragmas (seems a little more "internal"). Do benchmarks show that these provide provide a significant enough improvement to justify these? 

I'd expect most of these simple methods would be inlined anyway, so `go:nosplit` likely won't have any impact,
```
./accumulator.go:43:6: can inline NewAccumulator
./accumulator.go:49:6: can inline NewAccumulatorWithSeed
./accumulator.go:55:6: can inline Accumulator.AddString
./accumulator.go:71:6: can inline Accumulator.AddUint64
./accumulator.go:87:6: can inline Accumulator.Value
```

mway: Fair point - I'd been looking at these in a vacuum. It's been a bit since I had done benchmarks on this but I don't think there's a lot of value in hinting about stack checks when they'll be inlined. Thanks!

PR #150, opened 2021-02-23
Enable configurable TTL-based scope/metric deallocation

mway:
Presently, scope and metric caches exist indefinitely. This creates a significant problem for applications dealing with a large number of metrics, particularly if the majority of metrics are emitted infrequently, or if dimensions are volatile (e.g. request-based metrics).

While this doesn't solve memory design issues within Tally, it at least affords the ability to configure if/when old memory should be reaped.

### DIFF
--- a/scope.go
+++ b/scope.go
@@ -85,10 +85,11 @@ type scope struct {
 	histograms      map[string]*histogram
 	histogramsSlice []*histogram
 	timers          map[string]*timer
-	// nb: deliberately skipping timersSlice as we report timers immediately,
-	// no buffering is involved.
+	timersSlice     []*timer
 
 	bucketCache map[uint64]bucketStorage
+	lastReport  time.Time
+	root        bool
 }
 
 // n.b. This function is used to uniquely identify a given set of buckets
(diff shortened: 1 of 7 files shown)


scope.go
```diff
@@ -209,48 +224,78 @@ func newRootScope(opts ScopeOptions, interval time.Duration) *scope {
...
 }
 
-func (s *scope) cachedReport() {
+func (s *scope) cachedReport() (reported bool) {
 	s.cm.RLock()
 	for _, counter := range s.countersSlice {
-		counter.cachedReport()
+		if rep := counter.cachedReport(); rep {
+			reported = true
+		}
```
vdarulis, line 269: `reported = reported || counter.cachedReport()` ?

vdarulis (approved):

scope.go
```diff
@@ -85,10 +85,11 @@ type scope struct {
 	histograms      map[string]*histogram
 	histogramsSlice []*histogram
 	timers          map[string]*timer
-	// nb: deliberately skipping timersSlice as we report timers immediately,
-	// no buffering is involved.
+	timersSlice     []*timer
 
 	bucketCache map[uint64]bucketStorage
+	lastReport  time.Time
```
vdarulis, line 91: initialize value with time.now() on subscope creation?
also may just store an `int64` instead, since we don't care about timezones/monotonicity/etc.

scope.go
mway: The report calls are required and stateful, so we'd need to do it the other way around, but yeah, I like the conciseness. Thanks!

scope.go
mway: Fair point - agreed. Thanks!

PR #156, opened 2021-05-01
Share allocated tags among cached metrics

mway:
This change:

1. Introduces string interning for allocated metrics, including names and tag keys/values
2. Shares tags among metrics where possible
3. Introduces tag caching
4. Reduces histogram bucket allocations by reporting using closures instead of heap-allocating the buckets themselves

### DIFF
--- a/m3/reporter.go
+++ b/m3/reporter.go
@@ -25,6 +25,7 @@ import (
 	"io"
 	"math"
 	"os"
+	"runtime"
 	"sort"
 	"strconv"
 	"sync"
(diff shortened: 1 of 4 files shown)


m3/reporter.go
```diff
@@ -109,16 +112,16 @@ type reporter struct {
...
+	pending         atomic.Uint64
 	resourcePool    *resourcePool
-	status          reporterStatus
-}
-
-type reporterStatus struct {
-	sync.RWMutex
-	closed bool
+	stringInterner  *cache.StringInterner
+	tagCache        *cache.TagCache
```
abliqo, line 123: Do we assume that tags will be the same during the lifetime of reporter so we don't need to expire/evict cached entries?

abliqo (approved):

m3/reporter.go
vdarulis: tally so far had no LRU eviction or anything for scopes, so all that stayed in memory forever, which is bad for people who tag their metrics with UUIDs.

mway: Yes and no - reporters are scope-agnostic, so subscopes can continue to cause reporters to allocate, but if we assume that there is at least 2 metrics on every subscope, this pulls ahead pretty much immediately. Right now scopes hold onto the memory that cached reporters allocate indefinitely, which means that as soon as you allocate more than one metric on the same scope, you're re-allocating the same effective storage for every subsequent metric. So while linear growth is still an (eventual) problem, this at least substantially raises the ceiling (and we already would have hit limits sooner without this because we'd have been holding on to the extra storage).

There's another PR open (not planning to merge it yet) where it adds TTL-based metric eviction, which would feed into this to also evict cached tags if there were no more places that it was used.

vdarulis (approved):

PR #157, opened 2021-05-01
Replace scope.status with atomic.Bool

mway:


abliqo (approved):

scope.go
```diff
@@ -508,18 +500,12 @@ func (s *scope) Snapshot() Snapshot {
 }
 
 func (s *scope) Close() error {
-	s.status.Lock()
```
vdarulis, line 511: nice, I've always thought the myriad attempts to make close not racy led to this huge mess.

vdarulis (approved):

PR #158, opened 2021-05-01
Enable indepent subscope closing through io.Closer

mway:
This change enables using the `io.Closer` interface of `tally.Scope` to close and free a given subscope. It does not change the `tally.Scope` API; users must use type conversion to use this for now.

Additionally, this introduces the following behavior:

- closing the root scope closes all scopes
- closing a subscope only affects that subscope

Since there is no real "scope hierarchy", `Close()` is not cascading (outside of the root scope).

### DIFF
--- a/scope.go
+++ b/scope.go
@@ -90,6 +90,7 @@ type scope struct {
 	closed      atomic.Bool
 	done        chan struct{}
 	wg          sync.WaitGroup
+	root        bool
 }
 
 // ScopeOptions is a set of options to construct a scope.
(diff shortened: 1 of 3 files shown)


vdarulis (approved):

scope.go
```diff
@@ -500,19 +502,54 @@ func (s *scope) Snapshot() Snapshot {
...
+	s.cm.Lock()
+	s.gm.Lock()
+	s.tm.Lock()
+	s.hm.Lock()
+	defer s.cm.Unlock()
+	defer s.gm.Unlock()
+	defer s.tm.Unlock()
+	defer s.hm.Unlock()
+
+	for k := range s.counters {
```
abliqo, line 533: Instead of iterating, why not `s.counter = make(map[string]struct{})` and let go runtime garbage-collect the references?

PR #159, opened 2021-05-01
Pool borrowed report-time histogram tags

mway:
Instead of allocating the temporary storage necessary to append histogram tags, borrow it from a pool.

### DIFF
--- a/m3/reporter.go
+++ b/m3/reporter.go
@@ -348,11 +348,10 @@ func (r *reporter) AllocateHistogram(
 	var (
 		_, isDuration = buckets.(tally.DurationBuckets)
 		bucketIDLen   = int(math.Max(
-			float64(len(strconv.Itoa(buckets.Len()))),
+			float64(ndigits(buckets.Len())),
 			float64(minMetricBucketIDTagLength),
 		))
-		bucketIDLenStr        = strconv.Itoa(bucketIDLen)
-		bucketIDFmt           = "%0" + bucketIDLenStr + "d"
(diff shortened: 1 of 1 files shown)


m3/reporter.go
```diff
@@ -542,15 +541,28 @@ func (r *reporter) Tagging() bool {
 
 func (r *reporter) process() {
 	var (
-		mets  = make([]m3thrift.Metric, 0, r.freeBytes/10)
-		bytes int32
+		extraTags = sync.Pool{
+			New: func() interface{} {
+				return make([]m3thrift.MetricTag, 0, 8)
```
vdarulis, line 546: shouldn't you return a ptr `make(*[]m3thrift.MetricTag, 0, 8)`, or does this work because escape analysis does not screw things up, unlike when the pool is declared outside (eg. at package level)

mway: Yeah, good catch. I'm almost positive this will fail escape analysis even in the current scope, because it's not returned serially after, they're returned en masse at at flush time, so as far as the compiler is concerned I can't imagine how it wouldn't escape. I'll double check. Slice pointer derefs can end up being faster anyway despite dereferencing (probably mostly due to duffcopy'ing the header).

mway: Looks like I was wrong; I ran a long running test and reported malloc deltas at every pool.Get and there are no allocations here. I guess it does pass the escape analysis, since it doesn't leave the outer scope of the flush loop function.

vdarulis (approved):

PR #169, opened 2021-09-11
Remove executable bit from thirdparty/*.go

creekorful:
Hello there!

While packaging your library on Debian I've noticed that the Go files inside thirdparty directory were flagged as executable. 
This PR remove the executable bit.

Cheers,


mway:
Thank you!

PR #171, opened 2021-11-05
Fix statsd dependency

cretz:
From v3.1.1 to v3.2.1, https://github.com/cactus/go-statsd-client half-way moved to go mod and made the `statsd/` subdirectory a module in itself (granted they should have been using `/v3` path, but that came with later versions).

So this means across minor versions there is a mismatch between the `https://github.com/cactus/go-statsd-client` module and the `https://github.com/cactus/go-statsd-client/statsd` module. If you had the `https://github.com/cactus/go-statsd-client/statsd` module at v3.2.1 (i.e. https://github.com/cactus/go-statsd-client/commit/cb0885a1018c) and the latest tally library in your go mod and did a `go mod tidy`, on go 1.16 you break and on go 1.17 you get:

```
github.com/cretz/tally-sample imports
        github.com/cactus/go-statsd-client/statsd loaded from github.com/cactus/go-statsd-client/statsd@v0.0.0-20200423205355-cb0885a1018c,
        but go 1.16 would fail to locate it:
        ambiguous import: found package github.com/cactus/go-statsd-client/statsd in multiple modules:
        github.com/cactus/go-statsd-client v3.1.1+incompatible (C:\Users\user\go\pkg\mod\github.com\cactus\go-statsd-client@v3.1.1+incompatible\statsd)
        github.com/cactus/go-statsd-client/statsd v0.0.0-20200423205355-cb0885a1018c (C:\Users\user\go\pkg\mod\github.com\cactus\go-statsd-client\statsd@v0.0.0-20200423205355-cb0885a1018c)

To proceed despite packages unresolved in go 1.16:
        go mod tidy -e
If reproducibility with go 1.16 is not needed:
        go mod tidy -compat=1.17
For other options, see:
        https://golang.org/doc/modules/pruning
```

This PR upgrades statsd to the latest tag before they started using `/v4` and `/v5` imports in order to solve this. There are other potential solutions such as upgrading statsd client all the way to `/v5` or making the `statsd/` folder in here its own module (i.e. having its own `go.mod`) and not infecting others with this old statsd dependency.

Can this be merged (or solved another way) and be tagged w/ a minor version? Thanks!

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/uber-go/tally/v4
 go 1.15
 
 require (
-	github.com/cactus/go-statsd-client v3.1.1+incompatible
+	github.com/cactus/go-statsd-client/statsd v0.0.0-20200423205355-cb0885a1018c
 	github.com/pkg/errors v0.9.1
 	github.com/prometheus/client_golang v1.11.0
 	github.com/prometheus/client_model v0.2.0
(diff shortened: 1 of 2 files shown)


FreeGuwop:
This is also impacting our team when attempting to build from non-Go 1.17 versions. Can we please merge this? This seems like the least risk approach. We already use the replace directive to make this change a build time. However, each team needs to do this.

cretz:
@mway - Thoughts?

FreeGuwop:
Ping @cretz and @mway 

@mway Can we please merge this in?

mway:
Sorry for the delay - thanks!

mway:
This is included in [v4.1.0](https://github.com/uber-go/tally/releases/tag/v4.1.0).

PR #172, opened 2021-11-19
sanitizing tags after cache lookup to reduce allocations and other optimizations

kishoresenji:
Tagged scopes sanitize the tags and create an instance of the scope and cache that scope. Sanitization involves allocating a map. 

The change is to lookup from the cache using the tags before sanitization and add to the cache after tag sanitization. Both keys (pre and post sanitization) are added to cache so that multiple Tagged scopes which map to the same sanitized scope is still preserved. 

Other optimizations are 1) converting a []byte to a string for lookup when the Tagged scoped is already cached and 2) using a specialized object pool for []string.

This reduces the allocations by **75%** (8 to 2) in the case of cached subscopes (which is the normal scenario) and decreases the allocations for non-cached subscopes (19 to 18).

If we exclude the map allocation while calling Tagged(), the allocations drop from 6 to 0 (**100% reduction**)

Before:
```
BenchmarkNameGenerationTagged-12            	47474730	        21.53 ns/op	       0 B/op	       0 allocs/op
BenchmarkScopeTaggedCachedSubscopes-12      	  921057	      1275 ns/op	     816 B/op	       8 allocs/op
BenchmarkScopeTaggedNoCachedSubscopes-12    	  437010	      3811 ns/op	    2362 B/op	      19 allocs/op
```

After: 

```
BenchmarkNameGenerationTagged-12            	46205017	        22.03 ns/op	       0 B/op	       0 allocs/op
BenchmarkScopeTaggedCachedSubscopes-12      	 1856272	       593.2 ns/op	     336 B/op	       2 allocs/op
BenchmarkScopeTaggedNoCachedSubscopes-12    	  380500	      3952 ns/op	    2347 B/op	      18 allocs/op
```


pool.go
```diff
@@ -61,3 +74,31 @@ func (p *ObjectPool) Put(obj interface{}) {
...
+	select {
+	case v = <-p.values:
+	default:
+		v = p.alloc()
+	}
+	return v
+}
+
+// Put releases a string slice back to the pool.
+func (p *StringSlicePool) Put(obj []string) {
```
kishoresenji, line 99: Interestingly Put(obj interface{}) with a []string escapes to heap but not with *bytes.Buffer

kishoresenji, line 99: Interestingly Put(obj interface{}) with a []string escapes to heap but not with *bytes.Buffer
That is why to save an alloc a specialization is done for []string

key_gen.go
```diff
@@ -57,17 +79,14 @@ func KeyForPrefixedStringMap(
...
 }
 
-// keyForPrefixedStringMaps generates a unique key for a prefix and a series
-// of maps containing tags.
-//
-// If a key occurs in multiple maps, keys on the right take precedence.
-func keyForPrefixedStringMaps(prefix string, maps ...map[string]string) string {
-	keys := keyGenPool.stringsPool.Get().([]string)
+func keyForPrefixedStringMapsAsKey(prefix string, maps ...map[string]string) key {
+	keys := keyGenPool.stringsPool.Get()
```
jakobgt, line 83:
```suggestion
	// fixed size slice will be stack allocated
	keys := make([]string, 0, 16)
```

Instead of a using a object-pool, could you use a stack-allocated string slice as in my suggestion? The initial size is e.g., 16 and if there are more than 16 keys, the `append` in line 86 will extend it (and thus causing an allocation). The good thing is that you don't have to remember to clean up the slice afterward and it is inherently thread-safe.

key_gen.go
```diff
@@ -57,17 +79,14 @@ func KeyForPrefixedStringMap(
...
 	for _, m := range maps {
 		for k := range m {
 			keys = append(keys, k)
 		}
 	}
+	// 1 allocation as sort.Interface escapes to heap
 	sort.Strings(keys)
 
 	buf := keyGenPool.bufferPool.Get().(*bytes.Buffer)
```
jakobgt, line 92:
```suggestion
	buf := make([]byte, 0, 256)
```
If you similarly use a stack-allocated byte-slice, you avoid the same problems as with the string slice (note the size 256 it just a suggestion). 

For the `WriteString` calls below you can use `buf = append(buf, prefix...)` to write out the content of the string to the byte slice in a non-allocating fashion.

key_gen.go
```diff
@@ -57,17 +79,14 @@ func KeyForPrefixedStringMap(
...
-func keyForPrefixedStringMaps(prefix string, maps ...map[string]string) string {
-	keys := keyGenPool.stringsPool.Get().([]string)
+func keyForPrefixedStringMapsAsKey(prefix string, maps ...map[string]string) key {
+	keys := keyGenPool.stringsPool.Get()
 	for _, m := range maps {
 		for k := range m {
 			keys = append(keys, k)
 		}
 	}
+	// 1 allocation as sort.Interface escapes to heap
```
jakobgt, line 89: Interesting, somewhat surprising, and a little bit disappointing. :-)  It will likely be helped by the generics in Go 1.18, where you don't need the boxing as much (hopefully).

scope_registry.go
```diff
@@ -86,19 +86,27 @@ func (r *scopeRegistry) Subscope(parent *scope, prefix string, tags map[string]s
 		return NoopScope.(*scope)
 	}
 
-	key := scopeRegistryKey(prefix, parent.tags, tags)
-
+	preSanitizeKey := keyForPrefixedStringMapsAsKey(prefix, parent.tags, tags)
 	r.mu.RLock()
-	if s, ok := r.lockedLookup(key); ok {
+	if s, ok := r.lockedLookup(preSanitizeKey.CastAsString()); ok {
 		r.mu.RUnlock()
```
jakobgt, line 92: Just a random thought and should not be fixed here, but would it make sense to use sync.Map here? It avoids the mutex locking, when the key is already present (and just use an atomic value lookup) and thus you don't need to remember the unlocking here.

scope_registry.go
```diff
@@ -86,19 +86,27 @@ func (r *scopeRegistry) Subscope(parent *scope, prefix string, tags map[string]s
 		return NoopScope.(*scope)
 	}
 
-	key := scopeRegistryKey(prefix, parent.tags, tags)
-
+	preSanitizeKey := keyForPrefixedStringMapsAsKey(prefix, parent.tags, tags)
```
jakobgt, line 89: If you were to inline the `keyForPrefixedStringMapsAsKey` you would not need the object pool for the `bytes.Buffer`, but it does make this function longer. I'll leave it up to you to try it out (and decide what to do).

scope_registry.go
```diff
@@ -86,19 +86,27 @@ func (r *scopeRegistry) Subscope(parent *scope, prefix string, tags map[string]s
...
-	if s, ok := r.lockedLookup(key); ok {
+	if s, ok := r.lockedLookup(preSanitizeKey.CastAsString()); ok {
 		r.mu.RUnlock()
+		preSanitizeKey.Release()
 		return s
 	}
 	r.mu.RUnlock()
 
+	preSanitizeKeyStr := preSanitizeKey.CopyAsString()
+	preSanitizeKey.Release()
```
jakobgt, line 99: If you decide to leave the `bufferPool` in place, I would encourage you to add a test that verifies the behavior under high concurrency. It increases our confidence that the objects are released correctly, even when we later want to change the code (e.g., to leverage the Go generics)

jakobgt (commented):
Sweet! Thanks for doing this. Could you test the stack-allocated slices to see if the code becomes easier to read (and also save on the channel interaction). It should make the code simpler as you don't have the object pools to worry about.

scope_registry.go
kishoresenji: yes, this was the reason I kept it as object pool instead of trying your suggestion. If I have to make byte slice on the stack, the method has to be inlined for it to work. However the same method is used from other places as well and it would need duplication of logic.

scope_registry.go
kishoresenji: +1. We could do this optimization as a follow up.

scope_registry.go
kishoresenji: yes, this was the reason I kept it as object pool instead of trying your suggestion. If I have to make byte slice on the stack, the method has to be inlined for it to work. However the same method is used from other places as well and it would need duplication of logic.

key_gen.go
kishoresenji: It is a good suggestion for keys as the string slice is not returned from the method. However unfortunately the stack allocated string slice escapes to heap as the sort method is called.

pool.go
vdarulis: Could you use a `sync.Pool` instead? We've been trying to move away from channel-based pools since these predate significant improvements in the runtime for `sync.Pool`.

You would have to actually use the pool directly instead of wrapping it, and storing `*[]string` slices (pointers _to_ slice header), and use it like:
```
buf := _stringPool.Get().(*[]string)

fpr {
	// do something - sanitize, etc...
	*buf = append(*buf, foobar)
}

*buf = (*buf)[:0]
_stringPool.Put(buf)
```

Note that if you cast `*[]string` back to `[]string` by doing `goodLookingBuffer := *buf` it can escape, because with Go you sometimes just can't have nice things.

key_gen.go
vdarulis: to avoid this in `m3collector` we actually use a simple insertion sort as it's better than an alloc+virtualized call as tag count is bounded by limits on metric size.

pool.go
kishoresenji: thanks for that hint. I used sync.Pool now and it works nicely. I only used it for key_gen.go only and ObjectPool is still being used by reporter.

key_gen.go
kishoresenji: I have removed the alloc with sorting by using a pointer to sort.StringSlice and using the sort.Sort() method.

key_gen.go
kishoresenji: Done. Using a stack allocated byte slice.

key_gen.go
```diff
@@ -57,78 +57,62 @@ func KeyForPrefixedStringMap(
...
 		for k := range m {
-			keys = append(keys, k)
+			*keys = append(*keys, k)
 		}
 	}
-	sort.Strings(keys)
 
-	buf := keyGenPool.bufferPool.Get().(*bytes.Buffer)
+	// using a pointer to implementation of sort.Interface does not perform an alloc
+	sort.Sort((*sort.StringSlice)(keys))
```
jakobgt, line 69: It seems that the Golang escape analysis has determined that the argument to `sort.Sort` escapes and thus, it seems that you cannot call `sort.Sort` with a stack-allocated value (as it has to escape). Therefore, I would recommend following @vdarulis suggestion of using an insertionsort on these. 

Using the insertionsort, I was also able to shave off ~16% of the CPU time:
```
# Command to test is 
$ go test -c -o keyPrefix.out ; ./keyPrefix.out -test.count=10 -test.run=asdfasdfasdf -test.bench="BenchmarkScopeTaggedCachedSubscopes" -test.benchmem -test.cpu 16 -test.memprofile=/home/user/mem.profile -test.cpuprofile=/home/user/cpu.profile  -test.mutexprofile=/home/user/mutex.profile -test.blockprofile=/home/user/block.profile > {before,after}.txt

$ benchstat before.txt after.txt                                                                                                      09:06 23/11
name                           old time/op    new time/op    delta
ScopeTaggedCachedSubscopes-16    1.06µs ± 2%    0.88µs ± 2%  -16.96%  (p=0.000 n=9+10)

name                           old alloc/op   new alloc/op   delta
ScopeTaggedCachedSubscopes-16      336B ± 0%      336B ± 0%     ~     (all equal)

name                           old allocs/op  new allocs/op  delta
ScopeTaggedCachedSubscopes-16      2.00 ± 0%      2.00 ± 0%     ~     (all equal)
```
(there were 10 runs and `before.txt` contains the numbers for this PR and `after.txt` contains the numbers for when the `[]string` value is stack-allocated and using an adapted form of https://www.golangprograms.com/golang-program-for-implementation-of-insertionsort.html) :

key_gen.go
jakobgt: Ahh sorry, just noticed your comment on the heap-allocation. 

For reference, diving even deeper into `sort.Sort` (which is called by `sort.Strings`), it seems that the escape analysis has determined that parameters to `sort.Sort` escapes:
```
./key_gen.go:61:2: keys escapes to heap:
./key_gen.go:61:2:   flow: {heap} = &keys:
./key_gen.go:61:2:     from &keys (address-of) at ./key_gen.go:69:32
./key_gen.go:61:2:     from (*sort.StringSlice)(&keys) (interface-converted) at ./key_gen.go:69:31
./key_gen.go:61:2:     from sort.Sort((*sort.StringSlice)(&keys)) (call parameter) at ./key_gen.go:69:11
```
as far as I can understand the `(call parameter)` means that the parameter escapes. Thus there is no way to call `sort.Sort` with a stack-allocated value.

scope_registry.go
```diff
@@ -86,19 +87,25 @@ func (r *scopeRegistry) Subscope(parent *scope, prefix string, tags map[string]s
 		return NoopScope.(*scope)
 	}
 
-	key := scopeRegistryKey(prefix, parent.tags, tags)
-
+	buf := keyForPrefixedStringMapsAsKey(make([]byte, 0, 128), prefix, parent.tags, tags)
 	r.mu.RLock()
-	if s, ok := r.lockedLookup(key); ok {
+	if s, ok := r.lockedLookup(*(*string)(unsafe.Pointer(&buf))); ok {
```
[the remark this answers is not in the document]
kishoresenji: done.

key_gen.go
kishoresenji: thanks. I have removed heap allocated []string and also used insertion sort (copied from go [sort](https://cs.opensource.google/go/go/+/refs/tags/go1.17.3:src/sort/sort.go;l=38) pkg)

key_gen.go
```diff
@@ -57,78 +57,62 @@ func KeyForPrefixedStringMap(
 	return keyForPrefixedStringMaps(prefix, stringMap)
 }
 
-// keyForPrefixedStringMaps generates a unique key for a prefix and a series
-// of maps containing tags.
-//
-// If a key occurs in multiple maps, keys on the right take precedence.
-func keyForPrefixedStringMaps(prefix string, maps ...map[string]string) string {
-	keys := keyGenPool.stringsPool.Get().([]string)
+func keyForPrefixedStringMapsAsKey(buf []byte, prefix string, maps ...map[string]string) []byte {
```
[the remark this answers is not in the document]
kishoresenji: done

key_gen.go
kishoresenji: yea thanks. I have removed sync.Pool and using stack allocated string slices and using insertionSort (copied from the std go sort pkg). sort.Sort() does do insertionSort internally when the len is <= 12

key_gen.go
```diff
@@ -84,51 +68,39 @@ func keyForPrefixedStringMaps(prefix string, maps ...map[string]string) string {
...
+	return string(keyForPrefixedStringMapsAsKey(make([]byte, 0, 128), prefix, maps...))
 }
 
-func (s *keyGenerationPool) release(b *bytes.Buffer, strs []string) {
-	b.Reset()
-	s.bufferPool.Put(b)
-
-	for i := range strs {
-		strs[i] = nilString
+func insertionSort(keys []string) {
```
jakobgt, line 99: Just to catch any future regressions, should we have a unit test for this function? If it is covered elsewhere that is also cool

jakobgt (approved):
Very nice! 50% reduction in CPU time, ~60% reduction in bytes allocated and 75% reduction in number of allocations. The compound effect will be big.

key_gen.go
kishoresenji: thanks for the suggestion. added a testcase.

scope_registry.go
```diff
@@ -86,19 +87,30 @@ func (r *scopeRegistry) Subscope(parent *scope, prefix string, tags map[string]s
 		return NoopScope.(*scope)
 	}
 
-	key := scopeRegistryKey(prefix, parent.tags, tags)
-
+	buf := keyForPrefixedStringMapsAsKey(make([]byte, 0, 128), prefix, parent.tags, tags)
```
vdarulis, line 90: IDs in practice can easily breach 256 bytes, buf of size 128 is going to escape quite often due to realloc in non-test cases.

vdarulis (approved):
lgtm with nit

scope_registry.go
kishoresenji: updated to 256. thanks.

PR #173, opened 2021-11-24
Bump default M3 packet size

vdarulis:
I don't know why packets were bounded to MTU before, but let's make the default sane and not tank performance for non-trivial metric counts, even if the users are not using a wrapper on top of Tally that does the right thing configuration-wise.


abliqo:
LGTM

abliqo (approved):

PR #180, opened 2022-06-21
Enable staticcheck, fix issues

abhinav:
Add and enable staticcheck to tally, and fix the following issues:

```
instrument/call.go:51:2: field scope is unused (U1000)
m3/reporter_test.go:647:2: should use copy() instead of a loop (S1001)
multi/reporter.go:263:2: field histograms is unused (U1000)
prometheus/reporter.go:154:2: field histograms is unused (U1000)
scope_benchmark_test.go:185:9: this result of append is never used, except maybe in other appends (SA4010)
scope_test.go:104:2: field scope is unused (U1000)
stats.go:198:2: should use copy() instead of a loop (S1001)
stats.go:435:2: field cachedValueBucket is unused (U1000)
stats.go:436:2: field cachedDurationBucket is unused (U1000)
statsd/example/statsd_main.go:36:18: statsd.NewBufferedClient is deprecated: This interface is "legacy", and it [..]
```

The following issues were ignored:

`m3/resource_pool.go:79:24: func (*resourcePool).releaseProto is unused (U1000)`:
There are `nolint:unused` clauses in the file surrounding that, but it's not
clear why.
I'll defer to owners to delete this.

`m3/reporter.go:590:20: argument should be pointer-like to avoid allocations (SA6002)`:
This issue is legitimate. The value of the sync.Pool is slightly reduced
because each slice-to-`interface{}` conversion costs an allocation.
The pool should hold and return `*[]m3thrift.MetricTag` objects to prevent that
allocation.
The fix is straightforward, but testing/benchmarking that specific case would
take more time than I have right now,
so I'm deferring to owners to fix this in a follow up.

---

Commits in this PR are individually reviewable.


mway (approved):
thank you!

PR #183, opened 2022-09-21
Possibility to use different tags format for statsd

cv65kr:
With the new version of `cactus/go-statsd-client` there is a possibility to use different tags format - https://github.com/cactus/go-statsd-client/blob/master/statsd/client_test.go#L97


cv65kr:
@mway any chance to take a look on that?

PR #184, opened 2022-10-05
Use multiple subscope buckets

cdvr1993:
We have an application that has a cpu limit of 30 cores. During the start of it we see a considerable increase in latency 20ms -> 100ms. After tracing it down, we found the source of the problem in the lock contention from the subscope. We tried this change and were able to remove any latency increase.

I am also adding a parallel benchmark so that it is easier to see the lock contention.

**Benchmarks**
Current
Screen Shot 2022-10-04 at 9 09 05 AM

After the change
Screen Shot 2022-10-04 at 9 07 56 AM


cdvr1993:
@vdarulis @mway @andrewmains12 could you take a look at this?

scope_benchmark_test.go
```diff
@@ -139,6 +141,75 @@ func BenchmarkScopeTaggedNoCachedSubscopes(b *testing.B) {
...
+		for pb.Next() {
+			if rand.Intn(100) < percentageCached {
+				root.Tagged(cachedMap)
+				continue
+			}
+
+			n := atomic.AddInt64(&index, 1)
+			value := strconv.Itoa(int(n))
+
+			root.Tagged(map[string]string{
```
andrewmains12, line 204: Shouldn't we be storing the result of this somewhere, so as to not have the code optimized away? 

> Before concluding I wanted to highlight that to be completely accurate, any benchmark should be careful to avoid compiler optimisations eliminating the function under test and artificially lowering the run time of the benchmark.

https://dave.cheney.net/2013/06/30/how-to-write-benchmarks-in-go

scope_registry.go
```diff
@@ -88,28 +113,33 @@ func (r *scopeRegistry) Subscope(parent *scope, prefix string, tags map[string]s
 	}
 
 	buf := keyForPrefixedStringMapsAsKey(make([]byte, 0, 256), prefix, parent.tags, tags)
-	r.mu.RLock()
+	h := maphash.Hash{}
+	h.SetSeed(r.seed)
+	_, _ = h.Write(buf)
+	ptrToSubscope := r.subscopes[h.Sum64()%uint64(len(r.subscopes))]
```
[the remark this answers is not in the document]
cdvr1993: I did this because there was a comment before from vytenis saying that I was actually getting the copy of a lock. So I wanted the name to show that this is a pointer.

scope_benchmark_test.go
cdvr1993: I could do that, but I actually checked this was being called through a cpu profile.

Screen Shot 2022-10-10 at 12 26 45 PM

Do you still want me to store it? I would need to also store in multiple buckets so that it doesn't become a bottleneck for the benchmark.

scope_registry.go
```diff
@@ -21,64 +21,89 @@
...
 import (
+	"hash/maphash"
+	"runtime"
 	"sync"
 	"unsafe"
 )
 
 var scopeRegistryKey = keyForPrefixedStringMaps
 
 type scopeRegistry struct {
```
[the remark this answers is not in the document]
cdvr1993: I could do that, but that would affect performance because you are suggesting that a Get() is going to compute the bucket hash and then the hash for the internal map.

So what of the function Subscope which does multiple gets and multiple sets? We would be repeating the bucket hash many times when we don't need to.

Although I could create a function that returns the pointer to the subscope, but I don't feel that is clean to have a public api to return an internal pointer.

scope_benchmark_test.go
andrewmains12: Fair point r.e. the parallelization--so long as it's not getting optimized away I think it's good. Add a comment to the code documenting what you just mentioned so that others don't come in with the same idea/comment.

scope_registry.go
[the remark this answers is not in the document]
andrewmains12: Eh sure, usually that's just enforced via the types but not too terrible as is.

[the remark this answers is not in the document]
vdarulis: I got confused initially, but thinking about it more, go vet would most certainly catch this.

scope_registry.go
```diff
@@ -138,15 +169,15 @@ func (r *scopeRegistry) Subscope(parent *scope, prefix string, tags map[string]s
 		bucketCache:     parent.bucketCache,
 		done:            make(chan struct{}),
 	}
```
vdarulis, line 171: actually, it looks like this could be done outside of lock completely, or at least partially - so all the allocs wouldn't be done under lock - this should yield another perf improvement as lock would be held for a shorter time

scope_registry.go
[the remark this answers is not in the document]
vdarulis: feel free to revert to previous naming :)

scope_registry.go
cdvr1993: If we do this, there is a potential case where many goroutines are trying to acquire the write lock for the same key, so the allocations would be done for all of them.

vdarulis: honestly, re-reading this method, I don't even know why I thought this could be done outside of lock ...

scope_registry.go
```diff
@@ -21,64 +21,90 @@
...
+	subscopes []*bucketScope
+}
+
+type bucketScope struct {
+	mu sync.RWMutex
+	s  map[string]*scope
 }
 
 func newScopeRegistry(root *scope) *scopeRegistry {
+	procs := runtime.GOMAXPROCS(-1)
```
vdarulis, line 45: could we wire this in as part of ScopeOptions and config?
With GOMAXPROCs as default.

cdvr1993:
@vdarulis done. The benchmark improved slightly:

Screen Shot 2022-10-13 at 10 08 47 AM

Note: I change the names of the benchmarks because they were not reflecting what was actually happening.

scope_registry.go
```diff
@@ -21,64 +21,96 @@
...
-	root      *scope
-	subscopes map[string]*scope
+	seed maphash.Seed
+	root *scope
+	// We need a subscope per GOPROC so that we can take advantage of all the cpu available to the application.
+	subscopes []*bucketScope
 }
 
-func newScopeRegistry(root *scope) *scopeRegistry {
+type bucketScope struct {
```
mway, line 39: nit: `scopeBucket` (it's a bucket, not a scope)

scope_registry.go
```diff
@@ -88,28 +120,33 @@ func (r *scopeRegistry) Subscope(parent *scope, prefix string, tags map[string]s
 	}
 
 	buf := keyForPrefixedStringMapsAsKey(make([]byte, 0, 256), prefix, parent.tags, tags)
-	r.mu.RLock()
+	h := maphash.Hash{}
```
mway, line 123: style nit: group these, and use default construction for zero values:

```go
var (
  buf = keyForPrefixedStringMapsAsKey(make([]byte, 0, 256), prefix, parent.tags, tags)
  h   maphash.Hash
)
```

scope_registry.go
```diff
@@ -88,28 +120,33 @@ func (r *scopeRegistry) Subscope(parent *scope, prefix string, tags map[string]s
...
+	h.SetSeed(r.seed)
+	_, _ = h.Write(buf)
+	subscopeBucket := r.subscopes[h.Sum64()%uint64(len(r.subscopes))]
```
mway, line 126: How much latency does this add (for serial calls)?

scope_registry.go
```diff
@@ -138,15 +175,15 @@ func (r *scopeRegistry) Subscope(parent *scope, prefix string, tags map[string]s
...
+	subscopeBucket.s[key] = subscope
+	if _, ok := r.lockedLookup(subscopeBucket, preSanitizeKey); !ok {
+		subscopeBucket.s[preSanitizeKey] = subscope
 	}
 	return subscope
 }
 
-func (r *scopeRegistry) lockedLookup(key string) (*scope, bool) {
-	ss, ok := r.subscopes[key]
+func (r *scopeRegistry) lockedLookup(subscopeBucket *bucketScope, key string) (*scope, bool) {
```
mway, line 185: nit: `lookupWithLock` to be consistent with other naming

mway (commented):
Nice! Left a cursory first pass, will take another tomorrow. Thanks for doing this!

scope.go
```diff
@@ -95,13 +95,14 @@ type scope struct {
...
-	DefaultBuckets  Buckets
-	SanitizeOptions *SanitizeOptions
+	Tags                     map[string]string
+	Prefix                   string
+	Reporter                 StatsReporter
+	CachedReporter           CachedStatsReporter
+	Separator                string
+	DefaultBuckets           Buckets
+	SanitizeOptions          *SanitizeOptions
+	RegistryConcurrencyLevel uint
```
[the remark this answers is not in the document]
vdarulis: I'm fine with not exposing it as a public option for now, I was just thinking that 99% of the time this is wrapped in an opinionated config, so very few users really configure it by hand.

We could just keep the option private for benchmarking etc.  and come up with a better name, like `RegistryConcurrencyLevel uint -> registryShardCount uint`

scope_registry.go
cdvr1993: I didn't touch the name, do you still want me to change this?

scope_registry.go
cdvr1993: Not very clear about this. In my summary I have added the benchmarks for the non-parallel versions.

You can see that:
- When cached, this new version shows an increase of ~60 ns.
- When not cached it shows an improvement of ~300 ns. So I would like to say it is negligible.

mway: I think we might even be able to do a bit better - 60ns is a pretty hard hit (I understand that the gain is notable in the parallel case, but we can't add a ton of serial latency either).

I'm mostly afk today, but I'll run some tests tonight/this weekend to see if we can get it closer to parity with the non-striped version for serial cached calls.

Part of the issue is that we need to account for various modes of updating metrics, like `scope.Counter("foo").Inc(1)` calls in a relatively tight loop. Granted users should be optimizing these themselves anyway, but they often don't, so we don't want to degrade very much compared to before this change.

cdvr1993: Actually, I reran multiple times the benchmarks and I'm not able to get to below 700ns anymore on the old version. Both versions now sit at ~750ns.

cdvr1993: The benchmark shows 0.81% spent on maphash out of 31.6% (total for this benchmark).

Screen Shot 2022-10-14 at 2 29 09 PM

cdvr1993: One thing to do to address this is to allow users to disable sharding. If the application only accesses metrics serially it makes sense to do that, but I would assume most applications access them in parallel.

mway (approved):
lgtm

andrewmains12 (approved):

PR #185, opened 2022-10-07
Terminate timeloop faster for m3

artms:
We quickly start and stop process (it is wrapper for another command) and timeloop potentially can delay process shutdown by up-to 100ms (_timeResolution). Instead of just sleeping we as well wait for signal that shutdown is in progress and quickly terminate loop.

### DIFF
--- a/m3/reporter.go
+++ b/m3/reporter.go
@@ -697,7 +697,7 @@ func (r *reporter) timeLoop() {
 		select {
 		case <-time.After(_timeResolution):
 		    break
-		case <-doneCh:
+		case <-donech:
 		    return
 		}
 	}
--- a/m3/reporter.go
+++ b/m3/reporter.go
@@ -116,7 +116,6 @@ type reporter struct {
 	commonTags      []m3thrift.MetricTag
 	done            atomic.Bool
 	donech          chan struct{}
-	timeLoopDoneCh  chan struct{}
 	freeBytes       int32
 	metCh           chan sizedMetric
 	now             atomic.Int64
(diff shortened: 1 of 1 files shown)


rabbbit (approved):
Thanks!

(you probably need another review though)

m3/reporter.go
```diff
@@ -694,7 +694,12 @@ func (r *reporter) reportInternalMetrics() {
 func (r *reporter) timeLoop() {
 	for !r.done.Load() {
 		r.now.Store(time.Now().UnixNano())
-		time.Sleep(_timeResolution)
+		select {
+		case <-time.After(_timeResolution):
```
prashantv, line 698: `time.After` allocates a new timer + channel on each call, from the docs:

> If efficiency is a concern, use NewTimer instead and call Timer.Stop if the timer is no longer needed.

Please use a timer instead.

artms: Replaced with NewTicker, we always wait for same duration.

artms:
Production tests fail with errors that timeLoop goroutine is leaking, investigating, I suspect that production is not always properly discarding tally

artms:
So it seems code is correct, indeed there are a lot of leaking goroutines because tally reporter is not properly closed. So this change potentially can introduce regression for tests which try to detect leaking goroutines.

m3/reporter.go
```diff
@@ -692,9 +692,16 @@ func (r *reporter) reportInternalMetrics() {
...
 
 func (r *reporter) timeLoop() {
+	t := time.NewTicker(_timeResolution)
+	defer t.Stop()
 	for !r.done.Load() {
 		r.now.Store(time.Now().UnixNano())
-		time.Sleep(_timeResolution)
+		select {
+		case <-t.C:
+			break
```
mway, line 701: nit: `return`

mway (commented):
Thanks! Just fix lint and then lgtm.

m3/reporter.go
artms: I really need to continue looping in loop and "breaking" out of loop in case of donech being closed

mway: Oops, yep. Thanks!

PR #187, opened 2022-11-03
Test concurrent writes to reporter

brawndou:
This gives us unit test coverage for parallel usecases across multiple subscopes.

### DIFF
--- a/scope_test.go
+++ b/scope_test.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2021 Uber Technologies, Inc.
+// Copyright (c) 2022 Uber Technologies, Inc.
 //
 // Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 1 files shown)


scope_test.go
```diff
@@ -578,6 +578,60 @@ func TestHistogramSharedBucketMetrics(t *testing.T) {
...
+	var (
+		r                = newTestStatsReporter()
+		wg               = &sync.WaitGroup{}
+		workerCount      = 20
+		scopeCount       = 4
+		countersPerScope = 4
+		counterIncrs     = 5000
+		rs               = newRootScope(
+			ScopeOptions{
+				Prefix:         "",
```
andrewmains12, line 591: Nit: just omit empty values; more idiomatic that way.

andrewmains12 (approved):

PR #189, opened 2022-11-17
Report Tag cache size as part of internal metrics

brawndou:
This helps service owners know if they are causing Tally to use too much memory

PR #197, opened 2023-01-14
Report Tag cache size as part of internal metrics (#189)

brawndou:
Additional internal metrics will be emitted:

- `tally.internal.num-tag-cache` will report the number of tags
- `tally.internal.counter-cardinality` will report the number of counters across all scopes
- `tally.internal.gauge-cardinality` will report the number of gauges across all scopes
- `tally.internal.histogram-cardinality` will report the number of histograms across all scopes

Other changes:
- refactored tests to use internal metric counts defined in variables


brawndou:
backporting #189 to v3 branch

Issue #198, opened 2023-01-17
Panic with prometheus reporter: "tally.internal.counter-cardinality" is not a valid metric name


vincentdaniel:
[This change](https://github.com/uber-go/tally/pull/189) included in release [v4.1.5](https://github.com/uber-go/tally/releases/tag/v4.1.5) introduced new metrics with names that seem not valid metric names for prometheus:

```
panic: descriptor Desc***fqName: "tally.internal.counter-cardinality", help: "tally.internal.counter-cardinality counter", constLabels: ***, variableLabels: [version]*** is invalid: "tally.internal.counter-cardinality" is not a valid metric name

goroutine 8 [running]:
github.com/uber-go/tally/v4/prometheus.NewReporter.func1(***0x1ff88a0, 0xc0004804e0***)
	/home/runner/go/pkg/mod/github.com/uber-go/tally/v4@v4.1.5/prometheus/reporter.go:306 +0x76
github.com/uber-go/tally/v4/prometheus.(*reporter).AllocateCounter(0xc0002f8e00, ***0x1d94d48, 0x22***, 0xc000310360)
	/home/runner/go/pkg/mod/github.com/uber-go/tally/v4@v4.1.5/prometheus/reporter.go:366 +0x322
github.com/uber-go/tally/v4.(*scopeRegistry).reportInternalMetrics(0xc000310780)
	/home/runner/go/pkg/mod/github.com/uber-go/tally/v4@v4.1.5/scope_registry.go:262 +0x6b5
github.com/uber-go/tally/v4.(*scopeRegistry).CachedReport(0xc000310780)
	/home/runner/go/pkg/mod/github.com/uber-go/tally/v4@v4.1.5/scope_registry.go:99 +0x91
github.com/uber-go/tally/v4.(*scope).reportRegistry(0xc00028d080)
	/home/runner/go/pkg/mod/github.com/uber-go/tally/v4@v4.1.5/scope.go:262 +0xd5
github.com/uber-go/tally/v4.(*scope).reportLoopRun(0xc00028d080)
	/home/runner/go/pkg/mod/github.com/uber-go/tally/v4@v4.1.5/scope.go:254 +0x59
github.com/uber-go/tally/v4.(*scope).reportLoop(0xc00028d080, 0x0?)
	/home/runner/go/pkg/mod/github.com/uber-go/tally/v4@v4.1.5/scope.go:242 +0xbb
github.com/uber-go/tally/v4.newRootScope.func1()
[...]

PR #203, opened 2023-01-26
fix broken internal metrics names and make them optional

brawndou:
This addresses https://github.com/uber-go/tally/issues/198

### DIFF
--- a/scope_registry.go
+++ b/scope_registry.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2022 Uber Technologies, Inc.
+// Copyright (c) 2023 Uber Technologies, Inc.
 //
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
@@ -34,9 +34,9 @@ var (
 
 	// Metrics related.
 	internalTags             = map[string]string{"version": Version}
-	counterCardinalityName   = "tally.internal.counter-cardinality"
-	gaugeCardinalityName     = "tally.internal.gauge-cardinality"
-	histogramCardinalityName = "tally.internal.histogram-cardinality"
+	counterCardinalityName   = "tally-internal-counter-cardinality"
+	gaugeCardinalityName     = "tally-internal-gauge-cardinality"
+	histogramCardinalityName = "tally-internal-histogram-cardinality"
 )
 
 type scopeRegistry struct {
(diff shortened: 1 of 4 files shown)


scope_registry.go
```diff
@@ -227,7 +231,7 @@ func (r *scopeRegistry) removeWithRLock(subscopeBucket *scopeBucket, key string)
 
 // Records internal Metrics' cardinalities.
 func (r *scopeRegistry) reportInternalMetrics() {
-	if r.skipInternalMetrics {
+	if r.internalMetricsOption != SendInternalMetrics {
```
brawndou, line 234: we can eventually set a default option if it's unset

scope_registry.go
```diff
@@ -34,9 +34,9 @@ var (
 
 	// Metrics related.
 	internalTags             = map[string]string{"version": Version}
-	counterCardinalityName   = "tally.internal.counter-cardinality"
-	gaugeCardinalityName     = "tally.internal.gauge-cardinality"
-	histogramCardinalityName = "tally.internal.histogram-cardinality"
+	counterCardinalityName   = "tally-internal-counter-cardinality"
```
SokolAndrey, line 37: The fact that this can cause a panic is not great. We should handle an invalid metric names gracefully. We could add sanitazation to the code where the metric is being registered. Though it probably makes sense to have it as a separate PR.

SokolAndrey, line 37: I don't think either dots or dashes are accepted by prometheus reporter. At least they not legal characters according to [Prom data model](https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels)

scope_registry.go
```diff
@@ -34,9 +34,9 @@ var (
...
+	gaugeCardinalityName     = "tally-internal-gauge-cardinality"
```
SokolAndrey, line 38: can we please have a test for that?
funny enough we have a comment in Prom reporter example about dashes and dots
https://github.com/uber-go/tally/blob/master/prometheus/example/prometheus_main.go#L37

scope_registry.go
brawndou: created https://github.com/uber-go/tally/issues/204 to track this request

brawndou: will change to underscore

scope_registry.go
brawndou: I'll have the scope's sanitizer process it during runtime instead. This way we can be agnostic about whatever scope/reporter/sanitizer is underneath.

PR #205, opened 2023-01-27
Fix for bad prometheus metrics name

brawndou:
This resolves https://github.com/uber-go/tally/issues/198 and improves usability of the internal metrics toggle and mirrors the change in the v3 branch.

### DIFF
--- a/scope_registry.go
+++ b/scope_registry.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2022 Uber Technologies, Inc.
+// Copyright (c) 2023 Uber Technologies, Inc.
 //
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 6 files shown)


scope_registry.go
```diff
@@ -228,40 +232,48 @@ func (r *scopeRegistry) removeWithRLock(subscopeBucket *scopeBucket, key string)
...
 	counters.Add(rootCounters.Load())
 	gauges.Add(rootGauges.Load())
 	histograms.Add(rootHistograms.Load())
 	log.Printf("counters: %v, gauges: %v, histograms: %v\n", counters.Load(), gauges.Load(), histograms.Load())
 
 	if r.root.reporter != nil {
-		r.root.reporter.ReportCounter(counterCardinalityName, internalTags, counters.Load())
-		r.root.reporter.ReportCounter(gaugeCardinalityName, internalTags, gauges.Load())
-		r.root.reporter.ReportCounter(histogramCardinalityName, internalTags, histograms.Load())
+		r.root.reporter.ReportCounter(r.root.sanitizer.Name(counterCardinalityName), internalTags, counters.Load())
```
SokolAndrey, line 262: this will execute sanitazation on each flush, right?

brawndou: yes, I'll refactor this code into the constructor so it's only run once

PR #212, opened 2023-02-17
fixing race condition in ForEachScope

brawndou:
This fixes an issue in ForEachScope() where a subscopeBucket could be updated during snapshot.

### DIFF
--- a/scope_registry.go
+++ b/scope_registry.go
@@ -125,11 +125,11 @@ func (r *scopeRegistry) CachedReport() {
 
 func (r *scopeRegistry) ForEachScope(f func(*scope)) {
 	for _, subscopeBucket := range r.subscopes {
+		subscopeBucket.mu.RLock()
 		for _, s := range subscopeBucket.s {
-			subscopeBucket.mu.RLock()
 			f(s)
-			subscopeBucket.mu.RUnlock()
 		}
+		subscopeBucket.mu.RUnlock()
 	}
 }
 
--- a/scope_registry_test.go
+++ b/scope_registry_test.go
@@ -25,7 +25,6 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
-	"github.com/stretchr/testify/require"
 )
 
 var (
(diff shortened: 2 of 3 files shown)


scope_registry_test.go
```diff
@@ -140,3 +141,21 @@ func TestNewTestStatsReporterManyScopes(t *testing.T) {
...
+	scope := NewTestScope("", nil)
+	go func() {
+		for {
+			hello := scope.Tagged(map[string]string{"a": "b"}).Counter("hello")
+			hello.Inc(1)
+		}
+	}()
+	var val CounterSnapshot
+	for {
+		val = scope.Snapshot().Counters()["hello+a=b"]
```
andrewmains12, line 155: This is kind of a nit, but should we express the test in terms of `ForEachScope` instead? Since that's where the actual bug is. Could do both tests if we're paranoid haha.

scope_registry.go
```diff
@@ -125,11 +125,11 @@ func (r *scopeRegistry) CachedReport() {
 
 func (r *scopeRegistry) ForEachScope(f func(*scope)) {
 	for _, subscopeBucket := range r.subscopes {
```
andrewmains12, line 127: Quick sanity check: r.subscopes is immutable after construction, right?

andrewmains12 (approved):

scope_registry.go
brawndou: yes, the buckets in r.subcsopes remain the same. though buckets will have subscopes added to them when subscope() is called.

scope_registry_test.go
brawndou: done, though ForEachScope()'s only real use case is snapshot() so the tests are pretty much the same

scope_registry_test.go
```diff
@@ -140,3 +140,29 @@ func TestNewTestStatsReporterManyScopes(t *testing.T) {
 		wantHistograms, r.counters[histogramCardinalityName].val,
 	)
 }
+
+func TestForEachScopeConcurrent(t *testing.T) {
+	root := newRootScope(ScopeOptions{Prefix: "", Tags: nil}, 0)
+	go func() {
```
rabbbit, line 146: Doesn't this leak goroutines during the test?

brawndou: good catch, fixing...

PR #214, opened 2023-02-25
Add goleak checking to tests

brawndou:
This adds goleak testing across all tests in package tally and fixes goroutine leaks in scope related tests.


scope_registry_test.go
```diff
@@ -143,10 +143,16 @@ func TestNewTestStatsReporterManyScopes(t *testing.T) {
...
-			hello.Inc(1)
+			select {
+			case <-quit:
+				return
+			default:
+				hello := root.Tagged(map[string]string{"a": "b"}).Counter("hello")
+				hello.Inc(1)
+			}
 		}
 	}()
```
mway, line 157: Even though we have partial synchronization here (send/recv across an unbuffered channel), it only covers the recv itself and the function could be preempted before exiting (particularly if more logic is added in the future), so we should still ensure that it exits before the test completes. Even just something like:

```go
var (
  root = newRootScope(/* ... */)
  quit = make(chan bool)
  done = make(chan struct{})
)

go func() {
  defer close(done)

  for {
    select {
    case <-quit:
      return
    default:
      // ...
    }
  }
}()

// ...

var c Counter
for {
  // ...
}

<-done
```

(Applies both here and the snapshot case below.)

scope_registry_test.go
```diff
@@ -143,10 +143,16 @@ func TestNewTestStatsReporterManyScopes(t *testing.T) {
 
 func TestForEachScopeConcurrent(t *testing.T) {
 	root := newRootScope(ScopeOptions{Prefix: "", Tags: nil}, 0)
+	quit := make(chan bool)
```
mway, line 146: nit: style

```suggestion
	var (
		root = newRootScope(ScopeOptions{Prefix: "", Tags: nil}, 0)
		quit = make(chan bool)
	)
```

scope_test.go
```diff
@@ -1167,16 +1172,23 @@ func TestSnapshot(t *testing.T) {
 
 func TestSnapshotConcurrent(t *testing.T) {
 	scope := NewTestScope("", nil)
+	quit := make(chan bool)
```
mway, line 1175: nit: style

```suggestion
	var (
		scope = NewTestScope("", nil)
		quit  = make(chan bool)
	)
```

scope_registry_test.go
brawndou: done

scope_registry_test.go
brawndou: done

scope_test.go
```diff
@@ -353,6 +354,10 @@ func (r *testStatsReporter) Flush() {
 	atomic.AddInt32(&r.flushes, 1)
 }
 
+func TestMain(m *testing.M) {
+	goleak.VerifyTestMain(m)
+}
```
mway, line 359: let's put this at the top of the file as the first test since it's `TestMain`

mway (approved):
lgtm modulo moving `TestMain` up top, thanks!

PR #218, opened 2023-05-15
Remove defer for improved inlining in v1.20+ using PGO

rajbarik:
Go [launched](https://go.dev/blog/go1.20) Profile-Guided Optimization (PGO) with v1.20. 
With this, inlining optimization now uses a threshold called `inlineHotMaxBudget` which has a default value of 2000 for PGO-based inlining [Line 76, [inl.go](https://github.com/golang/go/blob/master/src/cmd/compile/internal/inline/inl.go)]. 

However, inlining does not work if a function/method contains `defer` statements [Line  619, [inl.go](https://github.com/golang/go/blob/master/src/cmd/compile/internal/inline/inl.go)]

This PR eliminates many `defer` constructs in `scope.go` and `scope_registry.go` in order to enable more inlining. For example, smaller functions such as `scope::counter(sanitizedName string)` and `scope::gauge(name string)` should be inlined after removing `defer`s in them.

I should also mention that I have gone to the extreme in removing many `defer`s. We should vet all the changes keeping in mind `panic` behaviors and make sure they are correct before merging this PR.

### DIFF
--- a/scope.go
+++ b/scope.go
@@ -188,8 +188,8 @@ func newRootScope(opts ScopeOptions, interval time.Duration) *scope {
 	if interval > 0 {
 		s.wg.Add(1)
 		go func() {
-			defer s.wg.Done()
 			s.reportLoop(interval)
+			s.wg.Done()
 		}()
 	}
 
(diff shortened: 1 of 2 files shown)


rajbarik:
@prateek FYI.

rabbbit:
Did you measure the expected efficiency wins from this in any way? In other words, is the extra brittleness worth it?

I was hoping that https://go-review.googlesource.com/c/go/+/479095 would have fixed this - but it looks like they only change the closure behavior, not `defer`.

Is a similar change to allow inlinfing functions with `defer`s not possible?

PR #220, opened 2023-05-23
[Tally Snapshot] -- add some functionality to histogram snapshot so t…

AlPal8:
…hat it is easier to test.

Add the number of values and durations in order to enable easier histogram testing

### DIFF
--- a/scope.go
+++ b/scope.go
@@ -660,6 +660,12 @@ type HistogramSnapshot interface {
 
 	// Durations returns the sample values by upper bound for a durationHistogram
 	Durations() map[time.Duration]int64
+
+	// NumValues returns the number of values which were emitted by this metric
+	NumValues() int64
+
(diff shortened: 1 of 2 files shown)


scope.go
```diff
@@ -660,6 +660,12 @@ type HistogramSnapshot interface {
 
 	// Durations returns the sample values by upper bound for a durationHistogram
 	Durations() map[time.Duration]int64
+
+	// NumValues returns the number of values which were emitted by this metric
+	NumValues() int64
```
prateek, line 665: The problem with doing this is HistogramSnapshot is an interface... i.e. adding a method to an interface is a breaking version change. 

This is one of the things about our current Tally API - we expose everything via interfaces, which is a right PITA. We can't land this like this. 

We can make a better testing story in the current version tho. Lets chat offline once

PR #221, opened 2023-06-02
[bug] Ensure that parent scopes do not return closed subscopes

mway:
This changes subscoping behavior to check the closed-ness of a cached subscope (if found) within the registry, and if it is closed, report and remove it immediately so that a new subscope can be created. The previous eventually-consistent behavior is also still intact: closed subscopes will either be pruned preemptively (at subscoping time, when that same subscope would be returned) or lazily (at reporting time).

The net effect of this change is that it is no longer possible to have the registry return a newly-closed, to-be-pruned subscope that the user would not know will become non-functional after the next reporting cycle. This is generally observed as metrics suddenly not being reported, particularly when doing application-level scope caching.

Before this change, it was possible to have subscope closure result in a defunct scope. For example, consider the following series of events:

```go
mytags := map[string]string{
  "hello": "world",
}

subscope := parent.Tagged(mytags)
subscope.Counter("foo").Inc(1)
subscope.(io.Closer).Close()
subscope = parent.Tagged(mytags)
subscope.Counter("foo").Inc(1)
```

Assuming the calls to `Scope.Close` (subscope) and `Scope.Tagged` (parent scope) happened without a reporting cycle between them, the subscope returned by calling `Scope.Tagged` on the parent scope would be closed, as the following would happen internally:

- counter `foo` allocated within `subscope`
- `subscope` closed
- `parent.Tagged` returns the still-cached, now-closed `subscope`
- changes to `subscope` (new metrics, value updates, etc) are valid until the next reporting cycle
- reporting cycle happens, `subscope` is removed from registry due to being closed
- `subscope` points to a live object that is no longer functional due to not being known to the registry

~I'd prefer to enable this by default (and remove the feature flag) since the previous behavior is incorrect, but added the flag just in case we're feeling paranoid.~ The flag has been removed, and this will be the default behavior (as it would never be desirable to have a defunct scope be returned).


scope_registry.go
```diff
@@ -152,23 +152,51 @@ func (r *scopeRegistry) Subscope(parent *scope, prefix string, tags map[string]s
...
+		}
+
+		switch {
+		case parent.reporter != nil:
+			s.report(parent.reporter)
+		case parent.cachedReporter != nil:
+			s.cachedReport()
+		}
+
+		tags = parent.copyAndSanitizeMap(tags)
```
devastating, line 174: Move these two lines out of the if/else block and get rid of else?

mway: Thanks - we still need to do some of this conditionally, but we can at least restructure this a bit so there's no duplication.

devastating (approved):
Some comments; mostly lgtm.

scope_registry_external_test.go
```diff
@@ -34,107 +34,83 @@ import (
...
-			subscope = root.Tagged(tags)
-			subscope.Counter("b").Inc(456)
-
-			requireClose(t, root)
-			wg.Wait()
-		})
-	}
+	wg.Add(2)
+
+	// Expect counter A to be reported when ReportOnSubscopeClose is
```
devastating, line 51: nit: no ReportOnSubscopeClose

devastating (approved):

andrewmains12 (approved):

scope_registry.go
```diff
@@ -152,23 +152,56 @@ func (r *scopeRegistry) Subscope(parent *scope, prefix string, tags map[string]s
...
 
-	// heap allocating the buf as a string to keep the key in the subscopes map
-	preSanitizeKey := string(buf)
 	tags = parent.copyAndSanitizeMap(tags)
-	key := scopeRegistryKey(prefix, parent.tags, tags)
+	sanitizedKey = scopeRegistryKey(prefix, parent.tags, tags)
+
+	// If a scope was found above but we didn't return, we need to remove the
+	// scope from both keys.
+	if ok {
```
andrewmains12, line 183: The control flow here is a bit hard to grok for me; how hard/worse would it be if we move this into the above `if ok` check?

mway: It's essentially either this or an `if/else` with duplicated tag/key sanitization in both branches.

We only want to call `copyAndSanitizeMap` and `scopeRegistryKey` after we've done the lookup above and only if we're not returning a valid scope, so this felt preferable to duplicating code.

If it's helpful, I can expand on the comment a bit, but I think this is probably ok as-is.

brawndou (approved):
lgtm

PR #228, opened 2023-09-28
Test with go >= 1.18

vdarulis:
Doesn't make sense to test with ancient Go runtimes - generics are too sweet,
and `m3db/m3` is on 1.18.


mway:
Thanks @vdarulis!

PR #231, opened 2023-10-03
add custom tags to internal metrics

shaan420:


m3/config.go
```diff
@@ -49,6 +49,10 @@ type Configuration struct {
 	// HistogramBucketTagPrecision is precision to use when formatting the metric tag
 	// with the histogram bucket bound values.
 	HistogramBucketTagPrecision uint `yaml:"histogramBucketTagPrecision"`
+
+	// CommonTagsInternal are tags that should be added to all internal metrics
+	// emitted by the reporter.
+	CommonTagsInternal map[string]string `yaml:"commonTagsInternal"`
```
prateek, line 55: can you call this `InternalTags` or something (i just want the prefix to start with `Internal` to make it clear

m3/reporter.go
```diff
@@ -288,6 +289,11 @@ func NewReporter(opts Options) (Reporter, error) {
 	internalTags := map[string]string{
 		"version": tally.Version,
 	}
+
+	for k, v := range opts.CommonTagsInternal {
```
prateek, line 293: can you ensure we don't set host/instance here too please (i think we might be inheriting those from common tags)

prateek (approved):
lgtm

PR #236, opened 2023-10-25
[DO NOT LAND] Process only changed metrics

shaan420:
PoC purposes only
Only applies to counters and gauges at the moment (not histograms)

### DIFF
--- a/scope.go
+++ b/scope.go
@@ -102,6 +102,11 @@ type scope struct {
 	done        chan struct{}
 	wg          sync.WaitGroup
 	root        bool
+
+	counterChangeNotifyCh chan *counter
+	gaugeChangeNotifyCh   chan *gauge
+	changedCounters       []*counter
+	changedGauges         []*gauge
 }
 
 // ScopeOptions is a set of options to construct a scope.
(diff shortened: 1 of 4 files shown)


scope.go
```diff
@@ -281,6 +286,39 @@ func (s *scope) reportRegistry() {
...
+		select {
+		case c := <-s.counterChangeNotifyCh:
+			counters = append(counters, c)
+		case g := <-s.gaugeChangeNotifyCh:
+			gauges = append(gauges, g)
+		case <-ticker.C:
+			s.reportChanges(counters, gauges)
+			s.cachedReporter.Flush()
+			// Reset the changed counters and gauges
+			counters = counters[:0]
```
prateek, line 305: you need to set the elements in the slice to the zero value

scope.go
```diff
@@ -281,6 +286,39 @@ func (s *scope) reportRegistry() {
...
+func (s *scope) processLoop(interval time.Duration) {
+	ticker := time.NewTicker(interval)
+	counters := make([]*counter, 0, _defaultInitialSliceSize)
+	gauges := make([]*gauge, 0, _defaultInitialSliceSize)
+
+	defer ticker.Stop()
+	for {
+		select {
+		case c := <-s.counterChangeNotifyCh:
+			counters = append(counters, c)
```
prateek, line 298: what happens if `c` is part of a subscope that's closed before you process it?

scope.go
```diff
@@ -166,23 +169,25 @@ func newRootScope(opts ScopeOptions, interval time.Duration) *scope {
...
+		gaugesSlice:           make([]*gauge, 0, _defaultInitialSliceSize),
+		histograms:            make(map[string]*histogram),
+		histogramsSlice:       make([]*histogram, 0, _defaultInitialSliceSize),
+		prefix:                sanitizer.Name(opts.Prefix),
+		reporter:              opts.Reporter,
+		sanitizer:             sanitizer,
+		separator:             sanitizer.Name(opts.Separator),
+		timers:                make(map[string]*timer),
+		root:                  true,
+		counterChangeNotifyCh: make(chan *counter, 1024),
```
prateek, line 189: hm few thoughts about this channel 
- any concerns about blocking once we hit it? 
- how did you arrive at this size?
- maybe this should be configurable too? (akin to https://github.com/uber-go/tally/blob/master/m3/reporter.go#L160 for eg)

scope.go
prateek: what we talked about -- https://go.dev/play/p/WIRovDFJhE5

scope.go
shaan420: As discussed, this will have the effect of flushing out any pending accumulated metrics which is desirable.

PR #258, opened 2024-04-12
Drop support for very old Go versions

vdarulis:
spring cleaning time


abliqo (approved):

.github/workflows/test.yml
```diff
@@ -13,32 +13,16 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        go: ["1.17.x", "1.18.x"]
-        include:
-        - go: 1.18.x
-          latest: true
+        go: ["stable", "oldstable"]
```
rabbbit, line 16: what is this magic.

vdarulis: just RTFM'ed https://github.com/actions/setup-go - `setup-go@v3` added these:

> If stable is provided, action will get the latest stable version from the [go-versions](https://github.com/actions/go-versions/blob/main/versions-manifest.json) repository manifest.
> 
> If oldstable is provided, when current release is 1.19.x, action will resolve version as 1.18.x, where x is the latest patch release.
>

vdarulis: I really don't miss updating some `.$CI.yaml` every damn release.

PR #260, opened 2024-06-25
Update README.md to use v4 in go get

alexbozhenko:
https://go.dev/ref/mod#module-path

The command that mentioned in the readme actually will download v3 of the package. Is it intentional?
```
$ go get -u github.com/uber-go/tally
...
go: added github.com/uber-go/tally v3.5.10+incompatible
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 Fast, buffered, hierarchical stats collection in Go.
 
 ## Installation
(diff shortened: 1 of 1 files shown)


alexbozhenko:
@mway Hi. Could you please take a look?

alexbozhenko:
@vdarulis maybe you can take a look?

Issue #262, opened 2024-08-05
Use standard library errors instead of deprecated github.com/pkg/errors


jetexe:
Package [github.com/pkg/errors](https://github.com/pkg/errors) was deprecated and archived three years ago. All functionality of that package was introduced in the standard library in Go 1.13.

I propose that we drop this unnecessary package and replace it with the standard implementation.

PR #263, opened 2024-08-05
Use standard library errors instead of github.com/pkg/errors

jetexe:
Package [github.com/pkg/errors](https://github.com/pkg/errors) was deprecated and archived three years ago. All functionality of that package was introduced in the standard library in Go 1.13.

Closes: #262


tarampampam:
Friendly ping @robskillington @mway

tarampampam:
Friendly ping @robskillington @mway [2], you might have missed this PR, and it's quite useful

Issue #264, opened 2025-03-04
Inconsistent Bucket Sizing Overflow Behavior on Different Architectures


rossb83:
The bucket creating methods such as [this one](https://github.com/uber-go/tally/blob/master/histogram.go#L348) should return an error or panic if the bucket sizes overflow. Overflow behavior is inconsistent between cpu architectures:

For inputs like
```
tally.ExponentialDurationBuckets(time.Second, 1.3, 90)
```
x86 -> The last two buckets become equal negatives `[..., -2562047h47m16.854775808s, -2562047h47m16.854775808s]`
ARM -> The last two buckets are equal positives `[..., 2562047h47m16.854775807s, 2562047h47m16.854775807s]`

Given customers are already living with this behavior we should be careful about the repair
* returning an error/panic could cause startup issues where there were none before
* silently detecting and removing overflows inline could create inconsistent bucket sizes 
* etc..

PR #266, opened 2025-03-18
Fix duration buckets overflow

roman-mazhut:
Fix for #264 
This PR addresses an overflow issue in `ExponentialDurationBuckets`, which can cause negative durations due to integer wrap-around when converting large values into `time.Duration`.

### Problem
Under the hood, `time.Duration` is represented as an `int64` nanosecond value. When the exponential scaling factor produces a number too large for `int64`, an overflow occurs, resulting in buckets with negative durations on some architectures.

Example:
```
tally.ExponentialDurationBuckets(time.Second, 1.3, 90)
```
Results in:
**x86**: `[..., -2562047h47m16.854775808s, -2562047h47m16.854775808s]`
**ARM**: `[..., 2562047h47m16.854775807s, 2562047h47m16.854775807s]`
(ARM seems to perform saturating calculations instead of wrapping.)

### Solution
This PR introduces a `safeFloat64ToInt64` function, which checks for overflows before converting to `time.Duration`. If an overflow is detected, it returns `math.MinInt64` or `math.MaxInt64`, preventing undefined behavior.

Now, `ExponentialDurationBuckets` will still return identical buckets in case of overflow, but the behavior will be consistent across different CPU architectures.

## Summary by CodeRabbit

- **New Features**
  - Enhanced the logic for generating duration buckets, improving precision and handling of extreme input values.

- **Bug Fixes**
  - Added tests to ensure that the system behaves correctly under maximum limit conditions, preventing unexpected errors.

- **Tests**
  - Introduced new test functions to validate the behavior of duration calculations and conversions, enhancing overall test coverage.

- **Chores**
  - Updated the workflow configuration to support testing across multiple operating systems, improving testing flexibility.

### DIFF
--- a/histogram.go
+++ b/histogram.go
@@ -307,8 +307,9 @@ func LinearDurationBuckets(start, width time.Duration, n int) (DurationBuckets, 
 		return nil, errBucketsCountNeedsGreaterThanZero
 	}
 	buckets := make([]time.Duration, n)
-	for i := range buckets {
-		buckets[i] = start + (time.Duration(i) * width)
+	buckets[0] = start
+	for i := 1; i < len(buckets); i++ {
+		buckets[i] = safeDurationSum(buckets[i-1], width)
 	}
 	return buckets, nil
 }
--- a/histogram.go
+++ b/histogram.go
@@ -368,7 +368,7 @@ func ExponentialDurationBuckets(start time.Duration, factor float64, n int) (Dur
 	curr := start
 	for i := range buckets {
 		buckets[i] = curr
-		curr = time.Duration(float64(curr) * factor)
+		curr = time.Duration(safeFloat64ToInt64(float64(curr) * factor))
 	}
 	return buckets, nil
 }
--- a/histogram_test.go
+++ b/histogram_test.go
@@ -29,6 +29,11 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+func parseDuration(s string) time.Duration {
+	duration, _ := time.ParseDuration(s)
+	return duration
+}
+
 func TestValueBucketsString(t *testing.T) {
 	result, err := LinearValueBuckets(1, 1, 3)
 	require.NoError(t, err)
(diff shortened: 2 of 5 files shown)


prateek (commented):
hey thanks for starting this off, few different things to change up 

- [ ] do the bounds check on all the bucket ctors (linear/exponential/etc) for both bucket types (duration/values)
- [x] turn on CI for both x86+arm -- https://docs.travis-ci.com/user/multi-cpu-architectures/ seems like it should be quite tractable

histogram_test.go
```diff
@@ -166,6 +166,14 @@ func TestMustMakeExponentialDurationBucketsPanicsOnBadFactor(t *testing.T) {
 	})
 }
 
+func TestMustMakeExponentialDurationBucketsOverflow(t *testing.T) {
```
prateek, line 169: add a test to capture the expected generated buckets on overflow

histogram.go
```diff
@@ -368,7 +368,7 @@ func ExponentialDurationBuckets(start time.Duration, factor float64, n int) (Dur
 	curr := start
 	for i := range buckets {
 		buckets[i] = curr
-		curr = time.Duration(float64(curr) * factor)
+		curr = time.Duration(safeFloat64ToInt64(float64(curr) * factor))
```
prateek, line 371: this can end up generating invalid buckets. e.g. if you're generating 10 buckets and bucket 7 is already overflowing, the last 3 buckets are going to be redundant/duplicates; this will cause issues on the metric consumption side (`histogramPercentile` in m3ql will complain about duplicate bucket ranges) 

i.e. we should probably pick a semantic in terms of either truncating to generate only valid, non-duplicate buckets OR we should consider raising a panic when there's invalid bucket definitions. not convinced which is better - but we should pick a path in this diff.

roman-mazhut: I thought about removing duplicate buckets, but my concern is that the third argument of the `ExponentialDurationBuckets` function is the number of buckets so that some clients may rely on that number.
Also, raising a panic is not a backward-compatible change.

histogram.go
```diff
@@ -368,7 +369,7 @@ func ExponentialDurationBuckets(start time.Duration, factor float64, n int) (Dur
...
```
prateek, line 372: options we talked about offline, consider the existing signature 
```
func ExponentialDurationBuckets(start time.Duration, factor float64, n int) (DurationBuckets, error) {
```

we could change this to
```
func ExponentialDurationBuckets(
  start time.Duration, 
  factor float64, 
  n int,
  opts ...BucketOpts,
) (DurationBuckets, error) {
...
}

// you can have opts for "autofix buckets" or "error on invalid bucket range/overflow/etc
```

prateek (approved):

PR #270, opened 2025-05-20
Implement Scope Pooling Mechanism

rahultejwani:
This PR implements a scope pooling mechanism to improve memory efficiency in high-cardinality metrics scenarios. Key changes include: - Efficient object reuse for high-cardinality metrics - LRU-based pool eviction - Automatic hit rate tracking - Performance optimizations with bytes.Buffer - Pre-allocated maps to reduce GC pressure. Benchmarks show 57% reduction in allocations and 64% reduction in memory usage.

### DIFF
--- a/scope.go
+++ b/scope.go
@@ -94,6 +94,9 @@ type scope struct {
 	root             bool
 	testScope        bool
 	noCacheSubscopes bool
+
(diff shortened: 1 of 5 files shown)


rahultejwani:
Created against wrong branch

PR #271, opened 2025-05-21
export Close() method for subscopes

emmanuel-chukwu:
Add ClosableScope interface to embed Scope and expose a Close() method to allow explicit cleanup of subscopes. This enables better resource management by allowing clients to:

- Clear metrics for unused subscopes

- Remove scopes from registry

- Release associated resources

The method is idempotent and thread-safe, returning an error if the scope is already closed.

### DIFF
--- a/types.go
+++ b/types.go
@@ -62,7 +62,24 @@ type Scope interface {
 
 	// Capabilities returns a description of metrics reporting capabilities.
 	Capabilities() Capabilities
-	
+}
+
+// ClosableScope extends Scope with the ability to close the scope
+// and create closable subscopes.
+//
+//	 IMPORTANT: When using Prometheus reporters, users must take care to
+//		not create metrics from both parent scopes and subscopes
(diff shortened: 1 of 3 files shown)


types.go
```diff
@@ -62,6 +62,12 @@ type Scope interface {
 
 	// Capabilities returns a description of metrics reporting capabilities.
 	Capabilities() Capabilities
+	
+	// Close closes the scope and releases any resources.
+	// For root scopes, this will also close the reporter.
+	// For subscopes, this will clear all metrics and remove the scope from the registry.
+	// Returns an error if the scope is already closed.
+	Close() error
```
prateek, line 87: can't do this - it's a breaking change to add a method to a go interface in semver terms

prateek (changes requested):

types.go
emmanuel-chukwu: I've introduced a new ClosableScope interface to embed Scope with a Close() method

PR #279, opened 2026-09-11
Add the native histogram metric types

rluong003:
**Stack 2 of 4.** Based on #278. The diff shown here is this commit only.

Adds the native histogram metric types. `Scope` does not expose them yet, that is the next PR.

### Values and durations are separate types

```go
type NativeValueHistogram interface {
	RecordValue(value float64)
}

type NativeDurationHistogram interface {
	RecordDuration(value time.Duration) // recorded as seconds
	Start() Stopwatch
}
```

So one distribution cannot mix the two. `Histogram` already guards this at runtime by checking the `htype` it derived from the declared buckets (`stats.go`). A native histogram has no declared buckets, so there was nothing for that guard to key off — and the failure it prevents is worse here. `RecordDuration(250ms)` records `0.25` and `RecordValue(250)` records `250`: same metric, 1000x apart, and the result is a plausible-looking wrong percentile rather than a visible count gap.

Both types are thin views over one shared accumulator, which exposes only an unexported `record(float64)`. That is load-bearing — the variants embed `*nativeHistogram`, so any exported `Record*` on the accumulator would be promoted into both and hand a caller back the API the split exists to withhold. `TestNativeHistogramVariantsWithholdEachOthersAPIs` fails if that happens.

`Start()` lands only on the duration variant; a stopwatch over a unitless histogram never made sense.

### The backend seam

```go
type NativeHistogramData interface {
	Update(value float64)
	Count() uint64
	Clear()
	MarshalBinary() ([]byte, error)
}
```

Tally calls `Update` and `MarshalBinary` and ships the bytes. It never inspects the payload. `DefaultNativeHistogramMaxBuckets = 160`, matching the OpenTelemetry exponential histogram default max size.

### drain clears on marshal error

`drain` clears the accumulator whether or not `MarshalBinary` succeeded.

Marshalling is deterministic and touches no I/O, so re-presenting the same samples cannot succeed. Worse, for the exponential-histogram backends this is aimed at, `Clear` is the only thing that resets the accumulator's scale — scale only ever goes down. Retaining on error therefore wedges the histogram permanently: it never marshals again, never reports, and says nothing about it. Losing one interval beats never reporting again.

The drain is a delta, not a cumulative total.

### Tests

14 tests on the type itself, no `Scope` references — the scope-level ones arrive in #280 with the code they exercise.

Both new behaviours are mutation-tested rather than trusted on a green run: giving the shared accumulator an exported `RecordDuration` fails the withholding test, and dropping the `Clear()` from the error path fails `TestNativeHistogramReportDiscardsSamplesOnMarshalError`.

`go build ./... && go test ./... && gofmt -l .` clean on this commit alone.


native_histogram.go
```diff
@@ -0,0 +1,198 @@
...
+func (h *nativeHistogram) RecordStopwatch(stopwatchStart time.Time) {
+	d := globalNow().Sub(stopwatchStart)
+	h.RecordDuration(d)
+}
+
+// drain marshals everything accumulated since the last drain and resets the
+// accumulator, so each payload is a delta covering one reporting interval
+// rather than the distribution's whole history.
+//
+// A marshal error leaves the accumulator intact, so the samples are retried on
```
adamjeanlaurent: >  A marshal error leaves the accumulator intact

Curious, would a Marshal error ever be transient? I think marshaling should be deterministic?

Do you think there is a risk if for some reason we have corrupt histograms that always fail to marshal, it could cause a memory leak because we keep them around?

native_histogram.go
```diff
@@ -0,0 +1,198 @@
...
+	data NativeHistogramData,
+	cachedNativeHistogram CachedNativeHistogram,
+) *nativeHistogram {
+	return &nativeHistogram{
+		data:                  data,
+		cachedNativeHistogram: cachedNativeHistogram,
+	}
+}
+
+func (h *nativeHistogram) RecordValue(value float64) {
```
adamjeanlaurent: Do you think we need any mechanism to ensure users don't call `RecordValue` and `RecordDuration` on the same histogram? 

I'm assuming with regular histograms we solve this because users either explicitly use value or duration buckets. 

Maybe we could has 2 different types of native hist with 2 different constructors?  `durationNativeHistogram` and `valueNativeHistogram` ? Similar to value buckets vs. duration buckets?

adamjeanlaurent (changes requested):
LMK what you think about duration vs. value

native_histogram.go
rluong003: yeah you're definitely right we need some sort of mechanism so that they don't mix usage of those APIs on the same histogram. I'll probably do something like 

```
RecordValue() {
if histogram.type != valueHistgoram {
return
   }
}
```

native_histogram.go
rluong003: no you're right here marshaling is deterministic. i think we should just Clear() on each drain, and not leave anything intact on errors

adamjeanlaurent (approved):

PR #280, opened 2026-09-11
Expose native histograms on Scope

rluong003:
**Stack 3 of 4.** Based on #279. The diff shown here is this commit only.

Wires the types from #279 into `Scope`. This is the commit that makes the feature reachable.

```go
NativeValueHistogram(name string) NativeValueHistogram
NativeDurationHistogram(name string) NativeDurationHistogram
```

Values and durations get separate lookup namespaces over one shared accumulator slice. Everything past the record call — draining, reporting, snapshotting — is identical, so only the create-and-find path needs to know which kind it is holding.

A name registered as both yields two distributions reporting under it, exactly as `Counter("x")` and `Gauge("x")` already do. Tally does not arbitrate cross-type name collisions and this does not add a new kind of them.

### ScopeOptions

Two new fields:

- `NativeHistogramFactory func(maxBuckets int) NativeHistogramData` — the only place a backend is ever supplied. Leave it nil and every native histogram on the scope is inert.
- `DefaultNativeHistogramMaxBuckets` — defaults to 160 when unset or `< 1`.

The bucket budget is configured on the scope rather than passed per call. It follows from the encoding and the cost an application has signed up for, which is a deployment-wide decision — and these objects are memoized, so a per-call budget was silently ignored on every call except whichever one happened to run first.

Both propagate to subscopes through `scope_registry.go`, so a factory set once at the root reaches everything beneath it. There is a test for that specifically, since the registry's sharded subscope construction is easy to get wrong.

### Mirrors the existing Histogram plumbing

Both methods follow `Histogram` step for step: lookup under `RLock`, take the write lock, double-check, allocate the cached handle when a `CachedStatsReporter` is present, store. The new `nhm` mutex joins the existing lock chain in `report`, `cachedReport`, `clearMetrics` and `Snapshot`.

`Snapshot` gains `NativeHistograms() map[string]NativeHistogramSnapshot`, exposing name, tags and sample count. Both variants land in the one map, the way `Histograms()` already carries value and duration histograms together. Not the payload — the encoding is the application's, so tally has nothing meaningful to hand back.

### Deliberately not included

No cardinality gauge in `reportInternalMetrics`. Adding one would put a permanently-zero series on every scope that never touches this feature, same argument as the m3 counter in #278.

### Tests

8 scope-level tests: the scope-level bucket budget, no-factory inertness, subscope inheritance, memoization, the two variants as separate metrics, and snapshot contents.

I mutation-tested the report wiring rather than trusting a green run — disabling the `nhm` block in `report()` makes `TestScopeNativeHistogram` fail, so the test is actually exercising the path.

`go build ./... && go test ./... && gofmt -l .` clean on this commit alone.


types.go
```diff
@@ -54,6 +54,16 @@ type Scope interface {
...
 
+	// NativeHistogram returns the NativeHistogram object corresponding to the
+	// name, holding at most maxBuckets buckets. Pass 0 to use the scope's
+	// configured default.
+	//
+	// Unlike Histogram, bucket boundaries are derived from the observations
+	// rather than declared up front, and the distribution is reported as a
+	// single serialized payload. Requires ScopeOptions.NativeHistogramFactory
+	// to be set; without it the metric accumulates but is never reported.
+	NativeHistogram(name string, maxBuckets int) NativeHistogram
```
adamjeanlaurent: LMK what you think, but my vote would be to not allow users to set max buckets. I think we should just use 160 as the default for everyone. I think the less configuration for users the better, and us controlling this on our side lets us find a good accuracy / performance tradeoff.

adamjeanlaurent (changes requested):
maxBuckets comment

types.go
rluong003: yeah totally makes sense to me, ill just default to 160

adamjeanlaurent (approved):

